PR #37, opened 2018-07-31
(fix): Update typings for immutable array helpers

MonsieurMan:
## PR Checklist

## PR Type
```
[x] Bugfix
```

Currently `push`, `unshift`, `splice` and `toggle` use any as input.  
I don't know if it is intended behavior but it seems wrong to me and makes for bad IDE integration.

## What is the new behavior?
They now take either `T` or  `...T[]` as input type.

```
[x] Yes
[ ] No
```

For users providing non `T` or `T[]` types, the code won't compile anymore which is a actually a good thing.

## Note
If this is indeed not intended behavior, I strongly recommend setting `noImplicitAny` to true, it would've maybe help catch this error at compile time.


NetanelBasal:
That's a good idea, thanks. I would merge your PRs in the upcoming release.

Issue #52, opened 2018-08-19
EntityStore should handle updating of entity ids


DmitryEfimenko:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

EntityStore does not handle updating entity's id key correctly

## Expected behavior
The following test should pass:
```ts
it('should update entity id', () => {
  store.add(new Todo({ id: 1 }));
  store.update(1, { id: 2 });
  expect(store.entities[2].id).toEqual(2);
  expect(store._value().ids).toEqual([2]);
});
```
## Minimal reproduction of the problem with instructions
I have a usecase of optimistic add/update functionality. The code looks like this:
`CategoriesService`:
```ts
addCategory(name: string) {
  const dirtyId = this.categoriesStore.optimisticAdd(name);

  this.api.addCategory(name).subscribe(
    res => this.categoriesStore.updateOptimistic(dirty, res.id),
    err => {
      console.error(err);
      this.categoriesStore.deleteOptimistic(dirty);
    }
  );
}
```
`CategoriesStore`:
```ts
optimisticAdd(name: string) {
  const category = { id: -1, name, dirty: uuid() };
  this.add(category);
  return category.dirty;
}

updateOptimistic(dirty: string, categoryId: number) {
  this.update(e => e.dirty === dirty, { id: categoryId });
}

deleteOptimistic(dirty: string) {
  this.remove(e => e.dirty === dirty);
}
```

## What is the motivation / use case for changing the behavior?
Akita should support updating entity ids to enable use cases like the one described above.

I already fixed issue locally and can send a PR if you don't have any concerns.

DmitryEfimenko:
I went ahead and submitted a PR. Please review when you get a moment

PR #53, opened 2018-08-19
fix(update): allow EntityStore updating entity id

DmitryEfimenko:
## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/datorama/akita/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #52 

## What is the new behavior?
EntityStore now allows updating entity id when updating a single entity.
Otherwise, it throws an appropriate error.

```
[ ] Yes
[X] No
```


akita/src/internal/crud.ts
```diff
@@ -69,16 +69,28 @@ export class CRUD {
     };
   }
 
-  _update<T extends EntityState>(state: T, ids: ID[], newStateOrFn: object | ((e: Readonly<any>) => object)): T {
+  _update<T extends EntityState>(state: T, ids: ID[], newStateOrFn: object | ((e: Readonly<any>) => object), idKey: string): T {
     const updatedEntities = {};
 
+    let isUpdatingIdKey = false;
```
NetanelBasal: This is a rare case; I'm wondering if it's better to create a dedicate method for this to keep the `update()` method clean. Something like:

```ts
this.store.updateEntityId(oldId, newId);
```

What do u think?

akita/src/internal/crud.ts
```diff
@@ -92,15 +104,25 @@ export class CRUD {
...
+      updatedEntities[idToUpdate] = newEntity;
+    }
+
+    let updatedIds = state.ids;
+    let stateEntities = state.entities;
+    if (isUpdatingIdKey) {
+      const id = ids[0];
+      const { [id]: any, ...rest } = state.entities;
+      stateEntities = rest;
+      updatedIds = [...state.ids.filter(current => current !== id), idToUpdate];
```
NetanelBasal: Notice that you are changing the order here. You should keep the entities order otherwise the view will change. Just replace the old id with the new one.

akita/src/internal/crud.ts
DmitryEfimenko: I understand the desire to keep the implementation of the method short and nice. Indeed, updating the id field of the entity is an edge case. However, I encourage you to think of customers of Akita. The smaller the Akita's API, the easier it is for the developers to learn it. I'd rather try and avoid introducing a new method on the EntityStore class to handle this edge case.

akita/src/internal/crud.ts
DmitryEfimenko: good catch!
Can't "just replace" the old id with new id because in dev mode the ids are frozen. I'll use `.map` to do that.

akita/src/internal/crud.ts
```diff
@@ -69,16 +69,28 @@ export class CRUD {
     };
   }
 
-  _update<T extends EntityState>(state: T, ids: ID[], newStateOrFn: object | ((e: Readonly<any>) => object)): T {
+  _update<T extends EntityState>(state: T, ids: ID[], newStateOrFn: object | ((e: Readonly<any>) => object), idKey: string): T {
```
shaharkazaz: idKey needs to be of type `ID`

akita/src/internal/crud.ts
```diff
@@ -92,15 +104,25 @@ export class CRUD {
...
       }
 
-      updatedEntities[id] = newEntity;
+      updatedEntities[idToUpdate] = newEntity;
+    }
+
+    let updatedIds = state.ids;
+    let stateEntities = state.entities;
+    if (isUpdatingIdKey) {
+      const id = ids[0];
```
shaharkazaz: you can destruct directly `const [id] = ids;`

akita/src/internal/crud.ts
```diff
@@ -69,16 +69,28 @@ export class CRUD {
...
+
     for (let i = 0; i < ids.length; i++) {
       const id = ids[i];
+      idToUpdate = id;
       assertEntityExists(id, state.entities);
 
       const oldEntity = state.entities[id];
       const newState = isFunction(newStateOrFn) ? newStateOrFn(oldEntity) : newStateOrFn;
 
+      if (newState.hasOwnProperty(idKey)) {
```
NetanelBasal: You can't check against `newState`. Your update method could look like this:
```ts
 this.store.update(1, entity => {
   return {
      ...entity,
      id: 5,
      completed: !entity.completed
   }
})
```

DmitryEfimenko: that is a valid point.
Just curious though, The following would achieve the same result, right?
```ts
this.store.update(1, entity => {
   return {
      completed: !entity.completed
   }
})
```

DmitryEfimenko: So I guess I also need to check if the value of id field has changed. Something like:
```ts
if (newState.hasOwnProperty(idKey) && newState[idKey] !== oldEntity[idKey]) {
```

NetanelBasal: Yes but you cant force it.

NetanelBasal: Yes. This seems right.

akita/src/internal/crud.ts
DmitryEfimenko: will do! Not used to this magic yet :)

oleersoy:
I wonder if it's better not to update an entities ID, but instead copy the entities state to a new entity, and assign a new ID.  Then delete the old entity, if we want to delete it.  We might want to keep it for historical purposes.

For example the ID is for a product and the product has a new version so we are assigning a new ID and copying the products info over into the new product.  If we just change the ID, then we have to make sure we also change the id of all orders / customers / etc. that reference that product.  If we did that the historical record would b somewhat inaccurate because now the reference entities refer to a product that is newer than the one that was originally referenced ...

If we create a new product and copy the old products information over, then we don't break the referential integrity of existing records ...

NetanelBasal:
@oleersoy That's not the only problem. As I mentioned earlier, let's say you already have a component that listenes to the old entity:

`this.book$ = query.selectEntity(oldEntity)`

When you are replacing the id, you'll break this subscription, as the id doesn't exist anymore. It's a tricky thing to accomplish.

DmitryEfimenko:
@oleersoy Great job thinking of potential use cases that this change might affect!

Please correct me if I'm wrong, but I think Akita does not enforce referential integrity in any way. It's up to a developer to make sure that entity references are kept in sync. In fact, the decision on whether the entity should be kept intact for historical reasons or updated with new id might depend on business logic. In case the business logic required keeping the record, the code could look like this:
```ts
const oldRecord = this.productsStore.entities[id];
const newRecord = { id: newId, ...oldRecord  };
this.productsStore.add(newRecord);
```
I think if we decide to make this logic a default, the entityStore might end up with a bunch of orphan entities and lead to confucion, or maybe even unexpected bugs.

DmitryEfimenko:
@NetanelBasal Yeah, the use case you're describing is tough. I'm not sure if there is a solution to that one that Akita can provide out of the box. I think it would have to be up to a developer to structure application components in such a way that this use case doesn't happen.

NetanelBasal:
I tend to agree with @oleersoy. EntityStore is like a table in a database. You will not replace the entity's id in a real database. I think it's more sense to **disallow** this behavior and enforce the user to remove the old one and create a new row if that's really what he wants.

I also think it's a **very** rare case, when you'll need to change entity's id.

DmitryEfimenko:
I see your point. The undo / redo functionality is indeed important. If that's the case, then I propose these changes:
1. The following if statement in the update logic
```ts
if (newState.hasOwnProperty(idKey) && newState[idKey] !== oldEntity[idKey]) {
```
should throw an error. Preferably this error should have a link to the docs website that explains why this is the case and what developer should do instead (@NetanelBasal could you please provide a link? It doesn't need to work right away. Just something to put in the code).

2. A new method on the API just like @NetanelBasal suggested originally: `this.store.updateEntityId(oldId, newId);`

I though that it would not even make sense to have this new method if all it would do is (1) create a copy of an old entity with new id and (2) remove the old entity. However, it's also important to preserve the order of ids - have the new id placed where the old id was. This justifies the new method.

We might as well now discuss the API for this method.
* `updateEntityId(oldId: ID, newId: ID)` - seems completly reasonable
* `updateEntityId(oldId: ID, newEntity: E)` - in case developer wants to also update additional props (it's often that server will send something like dateModified info or whatever else calculated on the server side).
* `updateEntityId(oldId: ID, newId: ID, options: IUpdateEntityIdOpts)` where `IUpdateEntityIdOpts` could have a flag to keep the old entity. Do we need that? WDYT?

Also, in dev mode, should this method result in two messages (one for remove and one for add) or in one new message? I'm referring to code like this:
```ts
isDev() && globalState.setAction({ type: 'Add Entity' });
```

oleersoy:
Or never update the entity id directly on the same entity instance.  In other words no changes.  Instead we do this.
1) Create a new entity
2) Copy the properties from the old entity excluding the id
3) Delete the old entity when appropriate

The benefits are:
- Stays in line with the Redux pattern
- State is mmutable so undo / redo remains consistent
- Referential integrity constraints are naturally obeyed
- It allows the dev to change the id if they want to
- It keeps the Akita API and implementation as simple as possible

@DmitryEfimenko do you see any drawbacks to this approach?

DmitryEfimenko:
@oleersoy I think we're talking about doing the same thing. Internally the `updateEntityId` method will do these exact steps you described (plus assuring the order of ids). Even now implementation I came up with does exactly that. It does not update the entity id directly (see [lines 107](https://github.com/datorama/akita/pull/53/files#diff-b03fb99b181832886d1d4c751f95e61fR107) and [114-115](https://github.com/datorama/akita/pull/53/files#diff-b03fb99b181832886d1d4c751f95e61fR114)).

Unless I completly misunderstand you guys. If that's the case, please provide more details

DmitryEfimenko:
Is it possible you're talking about calling `setState` twice? First to add new entity and second to delete old? If that's the case, I'm wondering if it will result in UI flickering

oleersoy:
@DmitryEfimenko It could be :).  I kind of scanned the thread so I might be lacking on the details.  In general I don't think Akita should have this method:

```
updateEntityId(oldId: ID, newId: ID)
```

Instead use the model factory to create a new instance of the entity.  For example suppose it's a `Todo`.  

```
todo:Todo = createTodo(currentTodo); // create new todo with current todo
store.add(todo); //new todo with new id
store.delete(currentTodo); //Delete current todo if necessary
```

So in other words make all the actions explicit.  The above method calls could have wrappers in case we want do disconnect subscribers, perform cascading deletes, etc.

DmitryEfimenko:
this will place the new todo at the end of the list. Please see the use case I'm trying to tackle in the issue #52.

Let's see what @NetanelBasal has to say.

oleersoy:
I looked at https://github.com/datorama/akita/issues/52

I'm not quite following what the business or semantic case for it is?  Could you please explain?

DmitryEfimenko:
It's an implementation of optimistic UI. Meaning, the UI will assume that add/update/delete operation was successful before the actual response from the server. This makes your website seem faster to the user than it actually is. [Here's one of the top search results](https://uxplanet.org/optimistic-1000-34d9eefe4c05) from the Google on the topic.

oleersoy:
Sound good but why does it need to update the `id` property?

DmitryEfimenko:
In case of an add, the id is unknown. It will be generated by the server. That's why there is this line of code:
```ts
const category = { id: -1, name, dirty: uuid() }; // notice id: -1
```
After successful add on the server, we get a response with an actual id and we need to patch entity in the store with that id.

oleersoy:
Ahh - I see what you mean now.  Hmmm - That makes it a little trickier, but hopefully not that much trickier.

What about first adding the item to the list that the UI is rendering, then sending it to the server, getting a response and adding it to the store (If no errors), but don't rerender the UI?

I know that that is probably not Akita's standard flow, but there's gotta be a simple way to engineer that?

DmitryEfimenko:
> but don't rerender the UI

Uff, that sounds like opening a whole other can of worms.
Let's see what @NetanelBasal says.

DmitryEfimenko:
@oleersoy There should be no methods telling Akita or Angular to update UI. it's Angular's responsibility to syncronize state with the UI. A bit of code is required to achieve this syncronization. Usually it's done like so:
You have a loop somewhere in the parent component (Products), where you are iterating over either (1) entities, or (2) the ids.
* In case 1 you pass the whole entity to a child component (Product) via `Input()`- so patching entity will propagate this change to the component via RxJS selector in the parent component.
* In case 2 you pass the id to the child component and this component will subscribe to the entity by id (the use-case Nataniel has mentioned). In that case, you may need to handle it like I have shown in [my previous comment](https://github.com/datorama/akita/pull/53#issuecomment-414430653). I haven't tested this solution. Some additional changes may be required. In most cases developer won't have to do that since the id won't change.
 ---

@NetanelBasal 
Having this conversation is quite helpful. It made me re-think the whole situation in depth.

Indeed EntityStore resembles a table in a relational database (which I love). It's got a primary key; select, update, delete methods in the Query classes - just like in SQL; Because of that, Akita allows us to do reactive inner/left joins via RxJs `combineLatest`... awesome!

However, it's not quite the same. It's not a database. It does not have a purpose of storing customer's data for years and having an ability to see a record's history a year later and be able to revert it. That's one of the primary reasons not to delete records with primary ids from the database.
The same goes for updating primary keys. You don't want to do that to assure referential integrity. However, the use-case I'm trying to solve for does not break referential integrity. It's one of these cases which are valid for the UI, but perhaps not valid for a database.
A relational database is usually quite far away (actually on the opposite end) from the UI, and as such, may not have the same business requirements.

Provided the reasoning above I think that my original approach (having this functionality in the `update()` method) is probably the correct one.

The `updateEntityId()` method implementation would basically be the same code I wrote to handle it in the current `update()` method, which makes me come back to the argument of thinking about customers and trying to avoid adding new methods to the API.

I don't think this implementation will hurt ability to undo/redo actions since it will be just a regular state change of the store.

I absolutely agree that updating id is very rare. In fact, I can't think of any other case when this would be required. We may consider placing a `console.warn` message in the dev mode if we notice that developer is trying to change id. WDYT?

However, optimistic UI is getting traction. Frameworks like Meteor or GraphQL Apollo Client have support for optimistic UI out of the box. Being able to easily tackle it using Akita is important.

Sorry for such a long post :/

oleersoy:
@DmitryEfimenko I like the real time thinking / optimistic concept.  The part I don't get though is this part:

> In case 2 you pass the id to the child component and this component will subscribe to the entity by id (the use-case Nataniel has mentioned). 

IIUC if we are subscribing to the entity by ID and we change the ID then we break the subscription?

oleersoy:
I wonder if we could eliminate the ID from Akita completely, such that we could patch it, but it's the same as any other property so it does not matter.

For example suppose we have a Todo instance `{title: "Add optimistic UI to Akita", description: "Use symbols instead of IDs"}`

Then when the Todo instance is added to Akita, akita creates a `Symbol` instance, and adds that to the Todo instance as a key, along with a UUID as the value.

```
var todoSym = Symbol('id');
var todo = {title: "Add optimistic UI to Akita", description: "Use symbols instead of IDs"};
todo[todoSym] = new UUID()
todoStore.add(todoSym, todo);

//If needed get the symbol from the store via the todo instance
Symbol s = todoStore.getSymbol(todo);
//Get Observable representing the Todo
Observable o = todoStore.get(s);
```

We can now ask the store for an `Observable` that corresponds to the `todoSym` and hook that up to the component.  If any of the component properties are patched, that's OK.  It can still be uniquely identified in the DOM via the `todoSym` and the corresponding UUID that is the value that it helps reference on the `Todo` instance.  Thoughts?

I have to admit I pulled this out of my a$$ somewhat, but I'm hoping that maybe it makes sense to those that are much smarter than I am.

NetanelBasal:
These are my conclusions: 

The case where you generate a "temporary" id on the client and replace it after saving the data on the server it's utterly valid case.

Akita should support this case. We have two options for that:
1. Replace the id of the current entity, without creating a new one.
2. Create new entity (mirror of the old one), assign the new id and remove the old one. Also keeping the order is essential.

I can't see anyhow Akita can help in a case we mention that the user already subscribes to the old entity. It should be the responsibility of the developer to take care to update the queries. 

Any thoughts?

DmitryEfimenko:
yeah, I'm not sure about the symbols. 

as for the case 2, I've put together a [demo](https://stackblitz.com/edit/akita-todos-app-nq2mh4) of that use-case. Whenever the id changes, it destroys the component associated with that id and creates a new one.

DmitryEfimenko:
@NetanelBasal it's my understanding that the current code implements option number 2:
> Create new entity (mirror of the old one), assign the new id and remove the old one. Also keeping the order is essential.

I'm not even sure how to implement option number one. Assigning the new id to the entity or using `.splice` method on ids does not work because they are frozen.

NetanelBasal:
@DmitryEfimenko you are right, updated the comment. I want to hear the opinion of @oleersoy

oleersoy:
I really gotta hit the sack soon, but a few thoughts.

Ideally the scenario supports referential integrity while being optimistic.  For example if we have to wait for the response (In order to get the new ID) before we hookup another entity to an entity that we added then the UI is not that responsive / optimistic enough.  

Suppose we add a library and then add a book to the library.  That should work optimistically with minimum effort on the developers part and should still be as responsive and performant as possible.

I think @DmitryEfimenko might be onto something with the Demo.  I need to read through the code more.  Looks really good so far though.

Ideally though we have something that offers and easy and error free developer workflow while guaranteeing referential integrity for scenarios that require it.  In other words we say just do A) B) C) and the UI is optimistic yet referential integrity is guaranteed.

I have a feeling that just assuming that it's OK to mutate the ID might cause errors / poor developer experience down the line, but I could be wrong.  If for example it was applied to the store demo with a card with products in it, and things could be added to a cart while keeping the UI optimistic and we all felt good that this would always be error free and easy to develop with then perhaps its a slam dunk.

oleersoy:
One conclusion that I think is fairly easy to reach is that we should try our best not to patch the primary key of the entity.

The minute we do that we have to worry about referential integrity.  For example suppose we are working offline in a PWA and we setup a a library with two books.  The books reference the library in a many to one relationship and the library references the books in a 1 to many relationship, so it's bidirectional.

Suppose we now go online and we want to sync our model graph with the server.  This should be really simple for the developer to do without having to worry about referential integrity or ids at all.

At this point I think that the simplest way to work with it is to assume that Akita has used something like a `Symbol` instance to construct and manage the graph and so it has not assigned or used the property that is the **primary key** of the entity.  Since it has not done that we (The developers) don't have to worry about it.  

Suppose we can only sync the library but not the books yet for some reason.  The library comes back with a primary key assignment and moves to a new model state, yet Akita still remembers which books belong to that library, even though the new primary key assigned to the library does exist as a foreign key on the books yet.

If this can be achieved then referential integrity of the domain in Akita lives outside of the referential integrity domain of the storage technology that is persisting the instances.  I think these two domains should be separate.

NetanelBasal:
The problem is that it's harder to manage symbols, for example, how you'll solve the problem with URLs that contain ids.

xxxtonixxx:
I think that @oleersoy 's idea is to use a symbol as akita internal ID, so you can modify the id of the record as any other field without alterate the order. The benefice of use symbol is that never will collide with other property of the record. Am I wrong, oleersoy?

oleersoy:
@xxxtonixxx  yes exactly.  It's for internal book keeping purposes only.  For example we could have a `Todo` instance created by the client factory, but not yet connected to a store.  
1) It's added to the store
2) The store adds a `Symbol('id')` key to the `Todo` that has a UUID value.

And that should be sufficient to create `Observable` instances from values in the store and it's cleanly separated from the backend.  The backend does not know anything about the `Symbol` parts of the model, and Akita does not care about the primary key of the entity instance.  That's left up to the client service / backend to manage as it sees fit.

oleersoy:
@NetanelBasal 

> Can you create a POC?

I certainly hope so.  I'm still reading through source and tutorials on Akita and solidifying my understanding of RxJS mechanics.

> Is this going to introduce breaking changes?

Hopefully not.  All we are doing is introducing another identifier for each model instance whose property key does not collide with any other keys and using that to create a permanent link between components and the model instance that the component is observing.  The rest should stay the same.  The end goal is just to decouple Akita's internal state management / communication from the state management that the client performs, so we are creating another layer, and not taking anything away. 

For example @DmitryEfimenko says that he prefers not to have to worry about both a generated primary key and a natural key.  This essentially creates a natural internal Akita key, so @DmitryEfimenko does not have to perform any additional work, but still gets the benefit of having a consistent identifier for all Akita instances.

NetanelBasal:
How you'll solve the problem with URLs that contain ids? for example - book/:id

DmitryEfimenko:
I think we might be overthinking the whole thing with the referential integrity. The optimistic UI case does not introduce an issue with referential integrity. It only requires patching id in one case: adding a record. Usually, it's a responsibility of a developer to disable any controls on the component that is optimistic and enable them once the server returned a successful response and patched the store entity with actual id.

I see the point @oleersoy is making. However, switching to use Symbols as an internal identifier for EntityStore would probably be a big breaking change. All the API (methods like update(), remove()...) would have to change to use that (Unless I don't quite understand @oleersoy's plan). It would also change the way a developer works with Akita. Imagine a simple case:
```ts
// component class
products$ = this.productsQuery.selectAll();

// template
<product *ngFor="let product of products$ | async" [product]="product" (updateName)="updateName(product, $event)"></product>

// products service
updateName(product: Product, newName: string) {
  const productSymbol =  Object.getOwnPropertySymbols(product)[0]; // that's super ugly, and there's an assamption that the symbol we're looking for is the first one in the array
  this.productsstore.update(productSymbol, { name: newName });
}
```
Again, perhaps I don't understand the whole idea. A POC would be helpful.

It's my understanding that the introduction of the symbol is an attempt to solve that one use-case where a component subscribes to an entity by id (the reference to the entity is lost when id changes). However, as my demo shows, this can be solved today. I also felt quite awkward having a subscription in child component. Even though it's possible someone will run into a situation like this, it's somewhat rare.

To summarize, I've slept on it, and I still think I'm right :)

NetanelBasal:
I'm with @DmitryEfimenko; I also can't found anyone mentioning this kinda of case googling for something like - "redux/ngrx replacing entity id." 

Akita can't be responsible for what the developers do. Akita can warn them, and that's all.

I think this is a very rare case and we are putting too much effort into it.

oleersoy:
So initally we have a new book:

```
{id: "", ISBN:"2353254-23523523-3252342", title: "Akita Rocks" }
```
Add it to the store (Adds the symbol and corresponding UUID and sets the id property):

```
{id: "UUID", ISBN:"2353254-23523523-3252342", title: "Akita Rocks", Symbol: UUID }
```
So now the instance has a unique ID, and it's being tracked by the symbol.  Later on we patch the ID:

```
{id: "354325", ISBN:"2353254-23523523-3252342", title: "Akita Rocks", Symbol: UUID }
```

So now it has a new ID, but Akita knows that this is still the instance identified by the [Symbol: UUID] key value pair.  Akita also only uses the [Symbol, UUID] key value pair for internal operations.
WDYT?

oleersoy:
> It's my understanding that the introduction of the symbol is an attempt to solve that one use-case where a component subscribes to an entity by id (the reference to the entity is lost when id changes).

@DmitryEfimenko it's trying to solve a lot more than that.  By decoupling state changes to the domain model from the internals of how Akita works we are potentially eliminating a lot of errors, and corresponding issue stack work down that comes from troubleshooting difficult scenarios...The types of scenarios that could also be expensive in terms of brand recognition and image for Akita ...  

I also think you are underestimating the referential integrity issues that may arise.  If we are saying that they could arise, but that's a dev error ... that's not good.  Because they will most likely show up on this issue stack and lead to a lot of wasted time on reengineering or lending support  that would not be necessary of the proper design is in place.

oleersoy:
> I'm with @DmitryEfimenko; I also can't found anyone mentioning this kinda of case googling for something like - "redux/ngrx replacing entity id."

@DmitryEfimenko might be right.  Maybe it's a non issue.  Could just try it and see.  If it does end up causing issues perhaps I'll have a proof of concept of the Symbol idea ready by then ...

DmitryEfimenko:
Sounds good to me

akita/src/internal/crud.ts
```diff
@@ -69,16 +69,28 @@ export class CRUD {
...
+
     for (let i = 0; i < ids.length; i++) {
       const id = ids[i];
+      idToUpdate = id;
       assertEntityExists(id, state.entities);
 
       const oldEntity = state.entities[id];
       const newState = isFunction(newStateOrFn) ? newStateOrFn(oldEntity) : newStateOrFn;
 
+      if (newState.hasOwnProperty(idKey) && newState[idKey] !== oldEntity[idKey]) {
```
NetanelBasal: I want to merge it. Can you add a test for this case, please? 

I mean the case when the id is present and the same.

DmitryEfimenko: sure thing. Will do it soon

oleersoy:
Just curious whether this might be something that also should be patched WRT the store implementation.

Suppose we have a Todo entity instance and three UI components have received an observable for it and are subscribing to the observables for change notification.  The Store changes the ID of the Todo instance along with the title.  The three subscribers now need to be updated.

Does the store update the key / ID used to lookup the observables when the ID for the Todo entity is changed.

I'm assuming the Observable instances that we need to call next on is looked up by the ID of the Todo entity instance?  So if the ID is patched, does the cache of observables that are being subscribed to also need to be updated?

NetanelBasal:
You should re-subscribe with the new id.

oleersoy:
@NetanelBasal  Do you know if there are any drawbacks to remapping the `Subject / Observable` so that the need to re-subscribe becomes unnecessary?

I'm trying to think what would happen if a store receives a push notification, and the `ID` is changed....the component does not necessarily realize this, but if the `Subject / Observable` can still notify it, then the design of the component is simpler... 

There's always the `Router / Navigation` by `ID` considerations, but if an `ID` change occurs via a push notification, then I'm assuming that the App needs to wrap the semantics of that in a service that understands what to do depending on the state of the application overall.

I'm thinking that there are two scenarios.  One in which the component is in a routed by `ID` view, so the router would have to renavigate the component if the `ID` changes, and another where it does not matter because the navigation URL does not have an `ID` in it....

oleersoy:
I build a POC of the Symbol idea and published it here (Note it mutates the state of the entities without storing undo / redo redux like history ... but does publish new arrays with the Observables):

https://github.com/fireflysemantics/slice

Also on NPM:

https://www.npmjs.com/package/@fireflysemantics/slice

It works by tagging each entity instance with a `GUID` (or UUID), such that each entity instance is globally identifiable from "Birth".  This has a lot of advantages when it comes to object comparison, etc.  

These tests show the general capabilities:
https://github.com/fireflysemantics/slice/blob/master/src/store.spec.ts

Basically you have a store that stores all the entities in general and you can add slices to the store keyed by an enum and observe the slices:
```
export const enum TodoSlices {
  COMPLETE = "Complete",
  INCOMPLETE = "Incomplete"
}

export class Todo {
  constructor(public complete: boolean, public title: string) {}
}

export let todos = [new Todo(false, "You complete me!"), new Todo(true, "You completed me!")];

let store: Store<Todo> = new Store<Todo>(todos);
store.addSlice(todo => todo.complete, TodoSlices.COMPLETE);

  let todos1$ = store.subscribe();
  let todos2$ = store.subscribe();

  it("should multicast the elements stored", () => {
    todos1$.subscribe(todos => {
      expect(todos.length).to.equal(2);
    });
    todos2$.subscribe(todos => {
      expect(todos.length).to.equal(2);
    });
  });```

It's still a bit raw, but I think it works in general.  Gotta start build prototypes with it now, to see if it truly has legs.  Curious what you guys think?

NetanelBasal:
Thanks. Will go over this soon.

Issue #125, opened 2018-11-22
feature : Plugin to manage filters


manudss:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior
have a better system for filters. There is already a filterBy function. When changing the filter, we must recall selectAll() to apply it. 

## Expected behavior
But it could be good to have a mechanic of multiple filter, via a filter bind.
As soon as we add a filter, or it is removed, it emit a new value filtered to all subscriber of select-all. 

## What is the motivation / use case for changing the behavior?
This could be useful for eg displaying products, like at amazon. We can filter the products with several filters applied. Then we can display the liste of alls filters, and remove one. 

This could also be usefull, to have global config. In config profil, we could set for exemple the liste of category you like it. And then, in all other page, this will display only data from these categories. 

The manage of filter is separated of the display, and any changed was apply in the list.

NetanelBasal:
As we discuss, it will be in a different package.

https://engineering.datorama.com/you-asked-for-it-we-listened-and-now-its-here-akita-v3-92740d0d72e4

tony:
@NetanelBasal Hi!   Any updates on this?  Where does this stand now?

NetanelBasal:
https://github.com/manudss/akita-filters-plugin

PR #126, opened 2018-11-22
Feature/filters plugin

manudss:
## PR Checklist

- [ x] The commit message follows our guidelines: https://github.com/datorama/akita/blob/master/CONTRIBUTING.md#commit
- [x ] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[x ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Have a better system for filters. There is already a filterBy function. When changing the filter, we must recall selectAll() to apply it. 

Issue Number: https://github.com/datorama/akita/issues/125

## Expected behavior
But it could be good to have a mechanic of multiple filter, via a filter bind.
As soon as we add a filter, or it is removed, it emit a new value filtered to all subscriber of select-all. 

This could be useful for eg displaying products, like at amazon. We can filter the products with several filters applied. Then we can display the liste of alls filters, and remove one. 

This could also be usefull, to have global config. In config profil, we could set for exemple the liste of category you like it. And then, in all other page, this will display only data from these categories. 

The manage of filter is separated of the display, and any changed was apply in the list. 

```
[ ] Yes
[x ] No
```


NetanelBasal:
Thanks, I appreciate the effort, but first, I have one crucial question.

Where do you see this functionality in use? In 99% of applications, we'll use a technique like pagination, and we'll not bring all the data in one time, so filtering on the client side isn't very useful as we want to filter the whole data and not just what is currently displayed. 

What I'm trying to say is that we need a server-side query and not client side.

manudss:
I am not agreed with you. Now with modern naigator, it could be easily to manage a large amont of data in client side. 

Obviously with a very large amount of data like Amazon, it will not be useful at all to make filters side client side. But for some data management applications, the local assets can better improve speed, and less calls the server.
Did you see, the playground example? I work with 200 data, and it works well. Working with such a quantity of data, is quite valid, without having to intervene server side.

I do not know where the 99% figure comes from? But it is true that having an option to manage server-side filters would also be a good idea. I think this feature could be add to pagination plugins because both can go peer.
But I'm sure having such a local version could be so useful:
For example :
- an app running in disconnected mode. (the offline mode can be managed by akita).
 - A store locator.
- A list of products (if there are not too many).
- The list of projects carried out by an agency.
- Results of a treatment,

I would like to put this in place on my app. Where I would manage the data in local. Hence the interest of creating this plug-in, because I tell myself that it can be used by others.

Most of your examples on docs or blogs use a filter.

I work for a computer service company, so I work on a lot of projects, and I have often had to work with data locally.

Of course, we may not need to have dozens of filters to apply. But 3 or 4 may be quite possible. Moreover, this approach makes it possible to define the filter management, of the display. in the example, I put the filters in a component, which works directly with the store filters.
And the twelfth component just has a subscribe on the data. By coding this example, I really found this approach interesting. Because the setting up of the filters was well facilitated. It also made it easy to store these filters. In the examples, if you change pages, and you come back, all the filters, are kept and restored.
Of course, all this can be done manually outside Akita, but I thought it could be added to Akita. And having it as a plug-in is a good idea. It is used only if it is needed.

When dealing locally data becoup, it's really possible. See an example of ag-grid, which displays a datagrid. It is able to handle 100,000 rows with 22 columns. And do all the processing (filters, grouping, search, etc.) are done on the client side.
https://www.ag-grid.com/example.php#/

So I think this plug-in can have its place.
And doing a version to handle server-side filters could be good too. To see how to do it!

NetanelBasal:
Ok, I will check the PR.

manudss:
> I see that you creating a `FiltersStore` inside the plugin, what if someone needs more than one plugin? currently you'll see one filters store in the devtools becuase they are set with the same name.

But the store will be different ? No ? 

Can we set another name ? That can be a mix of two name ?

NetanelBasal:
Yes, the store will be different, I guess we'll need to ask the user to provide custom name if he needs more than one plugin.

playground/src/app/products-filters/state/products-filters.service.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+
+    return this.productsQuery.isPristine ? request : noop();
+  }
+
+  setFilter(filter: Filter) {
+    this.filtersProduct.setFilter(filter);
+  }
+
+  setOrder(by: any, order: '+'|'-') {
```
NetanelBasal: setOrder => setSortBy

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,166 @@
...
+    return this._filterStore;
+  }
+
+  get filterQuery(): FiltersQuery {
+    return this._filterQuery;
+  }
+
+
+
+  constructor(protected query: QueryEntity<S, E>, private params: FiltersParams = {}) {
```
NetanelBasal: In a second thought, we must ask the user to provide us with the FiltersStore name as you can have multiple plugins. If the user doesn't provide a custom name, we can take the default Store name and set the name to `${this.store.name}Filters`

akita/__tests__/filters-plugin.spec.ts
```diff
@@ -0,0 +1,291 @@
+import { skip, take } from 'rxjs/operators';
+import { createWidget, createWidgetCompleted, WidgetsQuery, WidgetsStore } from './setup';
```
NetanelBasal: Please apply prettier to all files.

akita/__tests__/filters-plugin.spec.ts
```diff
@@ -0,0 +1,291 @@
...
+
+        filters.setFilter({id: 'filter3', hide: true, name: 'Filter 3', order: 2});
+        expect(filters.filterQuery.getCount()).toEqual(3);
+        expect(filters.filterQuery.getEntity('filter3').id).toEqual('filter3');
+        expect(filters.filterQuery.getEntity('filter3').hide).toEqual(true);
+        expect(filters.filterQuery.getEntity('filter3').name).toEqual('Filter 3');
+        expect(filters.filterQuery.getEntity('filter3').order).toEqual(2);
+
+
+        filters.setFilter({id: 'filter2', value: 'aaaa'});
```
NetanelBasal: each `it` should begin with the word `should`

akita/src/plugins/filters/filter-plugin.ts
```diff
@@ -0,0 +1,14 @@
+import { AkitaPlugin, Queries } from '../plugin';
+
+
+export class FilterPlugin<E = any, S = any> extends AkitaPlugin<E, S> {
```
NetanelBasal: Remove the file

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,166 @@
...
+  entityIds?: ID | ID[];
+  [key: string]: any;
+};
+
+/**
+ * Helper function to do search
+ * @param searchKey
+ * @param inObj
+ */
+export var searchObjFunction = function(searchKey: string, inObj) {
```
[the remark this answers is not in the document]
manudss: It is juste an helper function. No need to use it. We could have other helper function

akita/src/plugins/filters/filter-plugin.ts
manudss: Do I need to have a FilterPlugins ? I did not really understood how it works  all together ? 

Because I need it for : 
`protected instantiatePlugin(id: ID): P {
    return new FilterPlugin(this.query) as P;
  }`

Or maybe juste add it, in bottom of filters-plugins.ts with no export, juste to be keep in local ? It is that ?

akita/src/plugins/filters/filters-plugin.ts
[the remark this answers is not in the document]
manudss: I have create two more helper function. The default will be used. 
`/**
 * Helper function to do do a default filter
 * @param searchKey
 * @param inObj
 */
export function defaultFilter(searchValue: any, inElement: any) {
  if (isObject(inElement)) {
    return searchFilter(searchValue, inElement);
  }
  return searchValue === inElement;
}

/**
 * Helper function to do search on all string element
 * @param searchKey
 * @param inObj
 */
export function searchFilter(searchKey: string, inObj: Object) {
  return Object.keys(inObj).some(function(key) {
    return typeof inObj[key] === 'string' && inObj[key].toLocaleLowerCase().includes(searchKey.toLocaleLowerCase());
  });
}

/**
 * Helper function to do search in one key of an object
 * @param searchKey
 * @param inObj
 */
export function searchFilterIn(searchKey: string, inObj: Object, inKey: string) {
  return typeof inObj[inKey] === 'string' && inObj[inKey].toLocaleLowerCase().includes(searchKey.toLocaleLowerCase());
}`

akita/__tests__/filters-plugin.spec.ts
manudss: I thought it was applied after each commit. But I applied Prettier.

akita/src/plugins/filters/filters-plugin.ts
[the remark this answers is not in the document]
manudss: Moved to filters-store.ts tto avoid warning Circular dependency if use one function by default.

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * Remove a Filter
+   * @param id
+   */
+  removeFilter(id: string) {
+    this.filterStore.remove(id);
+  }
+
+  /**
```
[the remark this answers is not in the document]
manudss: I have seen, that reset must be enabled ? I am wonder why did you make this ? 

I have add a clean method : But maybe this would be in the entites stores class, no ? a clean method that remove all entites (or removeAll() ). 

`clean() {
    this.remove(this._value().id);
  }`

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,164 @@
...
+  setSortBy(order: SortByOptions<E>) {
+    this.filterStore.updateRoot({ sort: order });
+  }
+
+  private applyFilters(entities: E[], filters: Filter[]) {
+    let entitiesFiltered: E[] = entities;
+    for (let filter of filters) {
+      if (filter.function) {
+        // @ts-ignore
+        entitiesFiltered = entitiesFiltered.filter<E>(filter.function);
```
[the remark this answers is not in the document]
manudss: Yes your right. I rewrite, with the function some, to stop, when one filter hide the element. 

`  private applyFilters(entities: E[], filters: Filter[]): E[] {
    if (filters.length === 0) return entities;
    return entities.filter((value: E, index: number, array: E[]) => {
      return !filters.some((filter: Filter) => {
        if (filter.function) {
          return !filter.function(value, index, array, filter);
        }
        return false;
      });
    });
  }`

playground/src/app/products-filters/filters-form/filters-form.component.html
```diff
@@ -0,0 +1,83 @@
...
+      <option value="-origin">Origin - DESC</option>
+      <option value="+familly">Family - ASC</option>
+      <option value="-familly">Family - DESC</option>
+    </select>
+  </div>
+</div>
+
+<div>Filter :
+
+  <div class="chip" *ngFor="let filter of (filters$ | async)" (click)="removeFilter(filter.id)">
```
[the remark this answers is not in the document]
manudss: Yes, i have corrected this bug in last commit.

akita/src/plugins/filters/filters-plugin.ts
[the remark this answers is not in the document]
NetanelBasal: You can use `this.getStore().remove();`

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,164 @@
...
+    return this._filterStore;
+  }
+
+  get filterQuery(): FiltersQuery {
+    return this._filterQuery;
+  }
+
+  constructor(protected query: QueryEntity<S, E>, private params: FiltersParams = {}) {
+    super(query, params.entityIds);
+    this.params = { ...{ filtersStoreName: query.__store__.storeName + '/filters' }, ...params };
```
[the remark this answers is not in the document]
manudss: Their is no getStore() function in EntityCollectionPlugin.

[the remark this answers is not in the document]
NetanelBasal: You have on the `super()`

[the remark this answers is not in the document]
manudss: Sorry,  super() call to EntityCollectionPlugin, not AkitaPlugin. And in this class, their is no function getStore() !? : https://github.com/datorama/akita/blob/master/akita/src/plugins/entity-collection-plugin.ts 
With intellisens, i have no function getStore(): https://i.imgur.com/WbXN0oo.jpg 

Or maybe I miss somethings !?

akita/__tests__/filters-plugin.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+      });
+
+      it('should take all filters by default', () => {
+        expect(filters.filterQuery.getCount()).toEqual(3);
+        expect(filters.filterQuery.hasEntity('filter1')).toEqual(true);
+        expect(filters.filterQuery.hasEntity('filter2')).toEqual(true);
+        expect(filters.filterQuery.hasEntity('filter3')).toEqual(true);
+      });
+
+      it('should add filter if not exist with all default value', () => {
```
NetanelBasal:
```suggestion
      it('should add filter if not exist with all default values', () => {
```

akita/__tests__/filters-plugin.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+      it('should call getFilters() and get only filters not hidden ', () => {
+        filters.setFilter({ id: 'filter2', hide: true });
+        const filtersAll = filters.getFilters();
+
+        expect(filtersAll.length).toEqual(2);
+        expect(filtersAll[0].id).toEqual('filter1');
+        expect(filtersAll[1].id).toEqual('filter3');
+      });
+    });
+    describe('SelectALL with Filters ', () => {
```
NetanelBasal:
```suggestion
    describe('SelectAll with Filters ', () => {
```

akita/__tests__/filters-plugin.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+          .subscribe(result => {
+            expect(Array.isArray(result)).toBeTruthy();
+            expect(result.length).toEqual(2);
+            expect(result[0].id).toEqual(1);
+            expect(result[1].id).toEqual(3);
+            done1();
+          });
+      });
+
+      it('should apply 2 filter if provided when select all', done2 => {
```
NetanelBasal: You don't need the `done` callbacks; everything is synchrony.

akita/__tests__/filters-plugin.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+            expect(Array.isArray(result2)).toBeTruthy();
+            expect(result2.length).toEqual(1);
+            expect(result2[0].id).toEqual(3);
+
+            done3();
+          });
+      });
+    });
+
+    describe('SelectALL when any change in filter, or entities', () => {
```
NetanelBasal:
```suggestion
    describe('SelectAll when any change in filter, or entities', () => {
```

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,163 @@
...
+    return this._filterStore;
+  }
+
+  get filterQuery(): FiltersQuery {
+    return this._filterQuery;
+  }
+
+  constructor(protected query: QueryEntity<S, E>, private params: FiltersParams = {}) {
+    super(query, params.entityIds);
+    this.params = { ...{ filtersStoreName: query.__store__.storeName + 'Filters' }, ...params };
```
NetanelBasal:
```suggestion
    this.params = { ...{ filtersStoreName: this.getStore().storeName + 'Filters' }, ...params };
```

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,163 @@
...
+        return false;
+      });
+    });
+  }
+
+  protected instantiatePlugin(id: ID): P {
+    return new FilterPlugin(this.query) as P;
+  }
+
+  destroy(id?: ID) {}
```
NetanelBasal: Call `clearFilters` inside the destroy.

NetanelBasal: remove the id param

akita/src/plugins/filters/filters-plugin.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+class FilterPlugin<E = any, S = any> extends AkitaPlugin<E, S> {
```
NetanelBasal: remove this

akita/src/plugins/filters/filters-utils.ts
```diff
@@ -0,0 +1,34 @@
+import { isDefined, isObject } from '../../internal/utils';
+import { Filter } from './filters-store';
+
+/**
+ * Helper function to do do a default filter
```
NetanelBasal:
```suggestion
 * Helper function to do a default filter
```

playground/src/app/products-filters/filters-form/filters-form.component.ts
```diff
@@ -0,0 +1,107 @@
...
+  ngOnInit() {
+    this.getallFiltersValues();
+
+   this.filterForm.controls.search.valueChanges.pipe(untilDestroyed(this)).subscribe((search: string) => {
+      if (search) {
+        this.productsService.setFilter({
+          id: 'search',
+          value: search,
+          order: 20,
+          name: '" ' + search + ' "',
```
NetanelBasal: Use template strings

playground/src/app/products-filters/state/products-filters.service.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   *
+   * @returns {Observable<ProductPlant[]>}
+   */
+  get(): Observable<ProductPlant[]> {
+    const request = this.productsDataService.get().pipe(
+      tap(response => {
+        this.productsStore.set(response);
+        // applyAction(
```
NetanelBasal: remove the comment, please

playground/src/app/products-filters/state/products-filters.store.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@angular/core';
```
NetanelBasal: We're almost at the end, but we need documentation. Do you think you can create a markdown file that documents how to use the plugin and the API it exposes? Don't worry about the English; I will help with that.

manudss: Ok, I will make à documentation.

And for the filter store name ?

NetanelBasal: I will do this when you finish everything.

PR #209, opened 2019-04-17
feat(store,storeconfig): config for deepFreeze function

### LINKED ISSUE #124: Akita deep freeze is conflicting with moment.js date formatting logic (notably with mat-datepicker)
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

When adding a `moment.js` date object to the store (in non production mode of course), the object gets frozen. This appears to be affecting the prototype (or some other global construct in moment), such that even a moment date object created completely outside of akita is broken when formatted.

I discovered this issue using a `mat-datepicker` with minimal default configuration and assigned a moment date value to it. Important: This date object hadn't even come from akita store, but I have other moment objects stored in the store. 

In `moment.js` there is the [following code](https://github.com/moment/moment/blob/2e2a5b35439665d4b0200143d808a7c26d6cd30f/moment.js) which is part of the date formatting logic 

        this._longDateFormat[key] = formatUpper.replace(/MMMM|MM|DD|dddd/g, function (val) {
            return val.slice(1);
        });

The above code is reached when formatting the moment with a lowercase format string (which is one of the values for `defaultLongDateFormat` such as L, LL, LLL). This occured for me by simply assigning a moment instance to a `mat-datepicker` which [uses a lowercase `l` value by default](https://github.com/angular/material2/blob/master/src/material-moment-adapter/adapter/moment-date-formats.ts). So in this instance the moment.js code above has `key == "l"` - and it then runs the regex on the format string for `L` and tries to assign it back onto `_longDateFormat` as a new key for future use.

The above line of code fails because `Object.isFrozen(this._longDateFormat) == true` and gives the extremely cryptic message `ERROR TypeError: Cannot add property l, object is not extensible`.

> I just realized that possibly trying to format a date for `l` *before* I initialize akita may be the solution - but wanted to pass this on and document it the best I can because it was very obscure and others will surely run into this sooner or later. 
> *Edit*: **Yes!**, adding `moment().format('l')` before akita is initialized will fix this problem`, but would really love to hear your input. 

## Expected behavior

I would expect to be able to put a moment object in the store and it not to break moment for use elsewhere :-)

I can come back to create a demo later if needed if the explanation above doesn't suffice.

Need to be able to put a moment date object in the store and it not break.

## Environment

<pre><code>
Angular version: X.Y.Z

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>
- Latest version of moment with latest angular.
- I also updated akita (having seen that a recent change was made to the freezing logic). It still has this issue.

Thanks

orangeswim:
Adds deepFreezeFunction in StoreConfigOptions and unit tests for feature
Can use custom deepFreeze for complex objects in store

Resolves #124

## PR Checklist

- [X ] The commit message follows our guidelines: https://github.com/datorama/akita/blob/master/CONTRIBUTING.md#commit
- [X ] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[X ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Objects in store are frozen with deepFreeze on _setState. Complex states or entities might break when frozen (like moment.js) during development.

Issue Number: #124

## What is the new behavior?
User can now add a custom deepFreeze function to dictate Freezing behavior during development.

```
[ ] Yes
[X ] No
```

## Other information
Also added a simple unit test for deepFreeze function. Where do I update docs?

### DIFF
--- a/akita/src/storeConfig.ts
+++ b/akita/src/storeConfig.ts
@@ -4,6 +4,7 @@ export type StoreConfigOptions = {
   idKey?: string;
   storeName?: string;
   cache?: { ttl: number };
+  deepFreezeFunction?: (o: any) => any;
 };
 
 export type UpdatableStoreConfigOptions = {
--- a/akita/src/storeConfig.ts
+++ b/akita/src/storeConfig.ts
@@ -4,7 +4,7 @@ export type StoreConfigOptions = {
   idKey?: string;
   storeName?: string;
   cache?: { ttl: number };
-  deepFreezeFunction?: (o: any) => any;
+  deepFreezeFn?: (o: any) => any;
 };
 
 export type UpdatableStoreConfigOptions = {
--- a/akita/__tests__/deepFreeze.spec.ts
+++ b/akita/__tests__/deepFreeze.spec.ts
@@ -54,7 +54,7 @@ function deepFreezeCustom(o: ComplexState) {
 
 @StoreConfig({
   name: 'complexState2',
-  deepFreezeFunction: deepFreezeCustom
+  deepFreezeFn: deepFreezeCustom
 })
 class ComplexStoreCustom extends Store<ComplexState> {
   constructor() {
--- a/akita/__tests__/deepFreeze.spec.ts
+++ b/akita/__tests__/deepFreeze.spec.ts
@@ -1,0 +1,79 @@
+import { Store } from '../src/store';
+import { StoreConfig } from '../src/storeConfig';
+import { deepFreeze } from '../src/deepFreeze';
+
+class SpecialObject {
+  specialString: string = 'special';
+  specialNumber: number = 2;
+
+  constructor(params: Partial<ComplexState>) {
+    Object.assign(this, params);
+  }
+}
+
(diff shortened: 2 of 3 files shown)


akita/src/storeConfig.ts
```diff
@@ -4,6 +4,7 @@ export type StoreConfigOptions = {
   idKey?: string;
   storeName?: string;
   cache?: { ttl: number };
+  deepFreezeFunction?: (o: any) => any;
```
NetanelBasal: Can you change it to `deepFreezeFn`, please?

PR #243, opened 2019-07-05
Implement OnDestroy

SpeedoPasanen:
```
[ x ] Bugfix
```

When component doesn't implement OnDestroy, ngOnDestroy is never called. Unsubscription doesn't happen, possible memory leak if someone copies this example.

## What is the new behavior?
Unsubscribes when the component is destroyed.

## Other information

Ngx-take-until-destroy does not work unless someone calls ngOnDestroy, hence OnDestroy must be implemented.


shaharkazaz:
If you look at the code you can see that the `ngOnDestroy` implementation does exist, this is just a missing interface.
There is no memory leak.

SpeedoPasanen:
> If you look at the code you can see that the `ngOnDestroy` implementation does exist, this is just a missing interface.
> There is no memory leak.

Yes.. Thank god my summer holiday started today, brain works no more. Of course ngOnDestroy does get called if it exists, even though the class hasn't been marked to implement it. Sorry & thanks =)

PR #251, opened 2019-07-17
chore(entity-service): add ng entity service

NetanelBasal:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/datorama/akita/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```


akita/src/entityService.ts
```diff
@@ -0,0 +1,9 @@
+import { EntityState, getEntityType, getIDType } from './types';
+import { Observable } from 'rxjs';
+
+export abstract class EntityService<S extends EntityState> {
```
GrandSchtroumpf: I thought this EntityService should have a getter for the resourceName and baseApi.

angular/playground/projects/ng-entity-service/src/lib/ng-entity-service-Notifier.ts
```diff
@@ -0,0 +1,38 @@
+import { Subject } from 'rxjs';
+import { Injectable } from '@angular/core';
+import { filter } from 'rxjs/operators';
+import { Msg } from './types';
+
+export enum HttpMethod {
```
GrandSchtroumpf: any reason not using a const enum ?

angular/playground/projects/ng-entity-service/src/lib/ng-entity.service.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    this.loader.dispatch({
+      method: HttpMethod.GET,
+      loading: true,
+      entityId: isSingle ? idOrConfig : null,
+      storeName: this.store.storeName
+    });
+
+    return this.http.get(url).pipe(
```
GrandSchtroumpf: Should you not put the options as second parameter?

GrandSchtroumpf: Also, maybe check somewhere that responseType isn't blob, arrayBuffer, ... to warn the dev that this is not supported.

akita/src/entityService.ts
NetanelBasal: Yes, I thought about it, and it's relevant only for REST APIs in general. We could always change it later.

angular/playground/projects/ng-entity-service/src/lib/ng-entity-service-Notifier.ts
NetanelBasal: There are not included in the bundle. Users will not be able to consume them.

angular/playground/projects/ng-entity-service/src/lib/ng-entity.service.ts
NetanelBasal: Yes, thanks, missed it. They can't change the `responseType`, and it will be in the docs, so I think it's redundant for now.

Issue #286, opened 2019-08-26
QueryEntity selectMany using auditTime(0)


FERNman:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

QueryEntity ```selectMany``` returns an Obseravle with ```auditTime(0)```, which appends the return event to the JS event loop. Plain ```select``` calls, on the other hand, return immediately, causing a delay between the two.

## Expected behavior

```selectMany``` should return immediately, just like ```select```.

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/github-xbhyct
When you navigate to the TODO page, ```selectMany``` and ```selectCount``` are called. Although both could (and should) return at the same time, ```selectMany``` returns one change cycle later (see the alert message).

## What is the motivation / use case for changing the behavior?

We are currently implementing virtual scrolling and are sending an HTTP request to get the first x entities and the total entity count. Both of these are then plugged into our component, which calls a method to load more entities as soon as the user scrolls near the bottom of the table and as long as the total entity count is bigger than the fetched entity count. When the total count gets set one change cycle earlier, the user is obviously near the end of the table (as there are no entities, but the count is already loaded) and the 'load more' lambda is triggered.

In general, I'd say this can lead to confusion as it's unexpected.

## Environment

<pre><code>
Angular version: 8.0.6

Browser:
- [x] Chrome (desktop) version 76
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 10.16.0  
- Platform:  Windows 

Others:

</code></pre>

NetanelBasal:
This is by design. Otherwise, the observable will fire per entity which doesn't want we want. We want to group them to one `next`.

FERNman:
Ah, I see... Well, that's unfortunate. Thanks for the reply though!

FERNman:
Ok so I played around with it a little bit more and I think something like the following could do the job:

```typescript
selectMany<R>(ids: IDType[]): Observable<EntityType[]>;
[...]

PR #287, opened 2019-08-27
refactor: queryEntity selectMany

FERNman:
SelectMany no longer uses observables for each entity, but listens to store changes directly.

## PR Checklist

  Not necessary

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #286 

## What is the new behavior?
`selectMany` now directly listens to changes of the state. By using `distinctUntilArrayChanged()`, it is still possible to only emit new values if the array content really changes. This way, we don't have to use `auditTime(0)`, which makes it behave like the other select methods.

```
[ ] Yes
[x] No
```

### DIFF
--- a/akita/src/queryEntity.ts
+++ b/akita/src/queryEntity.ts
@@ -17,6 +17,7 @@ import { EntityAction, EntityActions } from './entityActions';
 import { isUndefined } from './isUndefined';
 import { QueryConfigOptions } from './queryConfig';
 import { distinctUntilArrayItemChanged } from './arrayFind';
+import { mapSkipUndefined } from './mapSkipUndefined';
 
 /**
  *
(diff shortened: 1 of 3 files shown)


akita/src/queryEntity.ts
```diff
@@ -129,12 +130,11 @@ export class QueryEntity<S extends EntityState, EntityType = getEntityType<S>, I
...
 
-    const entities = ids.map(id => this.selectEntity(id, project));
-
-    return combineLatest(entities).pipe(
-      map(v => v.filter(Boolean)),
-      auditTime(0)
-    );
+    return this.select(state => state.entities).pipe(
+      map(entities => ids.map(id => getEntity(id, project)(entities))),
+      map((entities): R[] | EntityType[] => entities.filter(Boolean)),
```
NetanelBasal: Thanks!

1. I know that it was there before, but we can save the second redundant iteration, and filter the `undefined` entities in the first loop.
2. Did you check that the typings are still working as expected?

FERNman: 1. I created a new helper method that uses a reducer to be able to skip entities without a second loop.
2. Yes, the typings are still correct. `getEntity` is the same as before, and `distinctUntilArrayItemChanged` will get either an array of type _R_ or _EntityType_, and will also return an array of the same type.

PR #294, opened 2019-09-12
feature(QOL): Allow the calling of upsertMany without being forced to…

AprilHickson:
… update the loading state

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently when you upsert many it will by default set the loading to false, this is frustrating when trying to store some values from another service into a not initialised store.

Issue Number: N/A

## What is the new behavior?
When passing in no options into the 'entityStore.upsertMany()' it will now not change the loading state unless a value is passed for options.loading

```
[x] Yes
[ ] No
```

if someone is relying on the loading to be updated by upsertMany they will have to change the calling to include the options of setting loading.

### DIFF
--- a/akita/src/entityStore.ts
+++ b/akita/src/entityStore.ts
@@ -269,7 +269,7 @@ export class EntityStore<S extends EntityState = any, EntityType = getEntityType
         ...state.entities,
         ...updatedEntities
       },
-      loading: !!options.loading
+      ...(options.loading == undefined ? undefined : { loading: options.loading })
     }));
 
     updatedIds.length && this.entityActions.next({ type: EntityActions.Update, ids: updatedIds });


NetanelBasal:
Thanks, but this is the desired behavior. In most cases, we'll call `upsertMany` after a server update. It should be aligned with the other methods such as `set`, and `add`. If you don't need it in your case, pass `loading: true`.

levitatejay:
Hi Netanel, here is our use case:

Entities:
```
Property {
   ...
}

Portfolio {
   ...
   properties: Property[]
}

Team {
   ...
   properties: Property[]
}

etc...
```
We use normalizr to flatten our API response and update the relevant stores.

During portfolioService.init we normalize and then upsertMany properties into the store however the propertyStore may or may not be already initialized. This is very common throughout our application as we lazy load different features.
i.e. we can have some (but not all) properties in the store as a result of initializing a different store.

Would you recommend something like the following:
```
const currentLoadingState = this.externalQuery.getValue().loading;
this.externalStore.upsertMany(entities, { loading: currentLoadingState })
```

NetanelBasal:
Yes, it looks good to me.

PR #388, opened 2020-03-15
refactor: moved repo to a nx monorepo structure

snebjorn:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[x] CI related changes
[x] Other... Please describe:
```

I had problems getting the local clone running and found a bunch of things that could be improved. Like upgrading typescript to > 3.2.4
This PR aims to make it super easy to get started while making it really hard to make unintended coding mistakes.

## What is the new behavior?

First off. We're running YARN instead. I believe this to be the right choice for a monorepo, especially v2 (berry). While these features technically aren't used, I believe the Nx tooling will move in that direction just as Lerna is.

Right... 2nd. We're using [Nx ](https://github.com/nrwl/nx) because it's a pretty cool monorepo tool that allow you to scaffold new apps and libs with builders/linters/testers setup and ready to go. BTW I've never used it before. And you can color me impressed.

Okay, new structure. I moved stuff around A LOT. The current structure was, sorry to say, a bit of a mess. The new structure is as follows

```
├── apps  # think examples
│   ├── angular-store-app
│   ├── angular-akita-server-side-pagination
│   ├── react-akita-app
│   └── ...
│
└── libs
    ├── akita
    ├── akita-ng-devtools
    ├── akita-ng-entity-service
    └── ...
```

Unfortunately I had some issue with the akita-schematics. see https://github.com/nrwl/nx/issues/2664
I wasn't able to make it automatically build and copy itself when akita was built using `nx`. ~So for now, and I know this sucks! You have to go into `libs/akita-schematics` and run `yarn build`.~ It's now taken care of when you build using `yarn build:lib:akita` 

The app/lib dependency graph looks like this

image

I moved the README and CHANGELOG for each lib into their respektive folders and setup standard-version to accommodate this. ~Unfortunately it doesn't completely play ball. So when generating changelogs you'll have to go into the library folder in question and run `standard-version` from there, either via global install or `npx`.~ Just run `yarn release:[library]`. Please checkout the `.versionrc` configs and see if you agree with it.

I've fixed all the applications so they actually run and use the latest version of angular/react. I didn't test them thoroughly. But they seem to work. I didn't touch the svelte app. Nx doesn't support that yet.

I've setup github actions to build, test and lint everything on PRs. So hopefully going forward it'll help with the code quality/safety.

I'm probably forgetting a lot. But I expect we'll have a lot to discuss :)

```
[ ] Yes
[🤞] No 
```

## Other information

I've removed this. It wouldn't compile and didn't seem to be used anywhere.
https://github.com/datorama/akita/blob/2f4d600c54428200df2d25a7b30b2ebc1e57c888/akita/src/root.ts#L3

There is a test I can't figure why is failing. It makes no sense!
https://github.com/datorama/akita/blob/2f4d600c54428200df2d25a7b30b2ebc1e57c888/akita/__tests__/pagination.spec.ts#L196
But the whole `pagination.spec.ts` suite is in dire needs of a complete re-write. ALL the tests in there bleed state to each other.

I replaced the guid() function (as you know) as it wouldn't build with the new version of ng-packagr.
https://github.com/datorama/akita/blob/2f4d600c54428200df2d25a7b30b2ebc1e57c888/akita/src/guid.ts#L11


libs/akita/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "version": "4.21.0",
+  "description": "State Management Tailored-Made for JS Applications",
+  "bugs": {
+    "url": "https://github.com/datorama/akita/issues"
+  },
+  "homepage": "https://github.com/datorama/tree/master/libs/akita#readme",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/datorama/akita",
+    "directory": "libs/akita"
```
snebjorn: I hope this setting does what I hope it does. Anyone that can tip in on that?

libs/akita-ng-entity-service/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "directory": "libs/akita-ng-entity-service"
+  },
+  "license": "Apache License 2.0",
+  "author": "Netanel Basal",
+  "keywords": [
+    "angular",
+    "akita"
+  ],
+  "peerDependencies": {
+    "@angular/common": ">= 9 < 10",
```
snebjorn: I was perhaps a bit too strict on this one. I believe the current version works with +7.

Also how do you perfer these ranges? Should we just say from vX and let users "discover" when it doesn't work anymore?
Or should we keep it as is `< 10` and let the users know that there are no guarantees. The latter requires a bit more work, but is also the better choice IMO

NetanelBasal: You can remove this. We're not coupled to Angular.

snebjorn: that package is

NetanelBasal: Angular plugins coupled to Angular.

snebjorn: Yes, and that's the package.json file for `@datorama/akita-ng-entity-service`

snebjorn: `>= 7 < 10` have been agreed upon

ntziolis:
@snebjorn First off we are also very happy nx users, but I want to provide some comments in regards to nx for publishable libraries:
- nx by design is using a single package.json file for all libs & apps
- yes its possible to create a publishable lib (node or angular) which then generates a  seperate package.json
- but there is no tooling support for those package.json files, they have to be maintained manuelly
- no integrated tooling support for semversioning
- nx makes heavy use of its affected feature
  - this is especially helpful to drive down build times in CI environments by only building what was affected since the last PR completed.
  - Making sure this works well with publishable libs reliably is not documented very well.  
     - Assuming we have libs A, B, C + D
     - B + C depends on A
     - We now want to build  B + C + D:
`nx run-many --target build --with-deps --projects "B,C,D" --prod`
- "-prod" flag when buidling
  - when creating angular publishable libs in nx there is a production config generated to disable Ivy (NG9)
  - when creating node libs there is no such config generated and using the --prod flag with those libraries will fail
  - we have resorted to simply creating production configurations (empty) for each lib that we are creating
  - this allows us to use one command to build them all while making use of the nx affected feature

Again, we are super happy with nx, but it is very apparent that the origins of nx is not in publishable packages.

Let me know if you have any questions, maybe we can assist.

snebjorn:
@ntziolis 
> * nx by design is using a single package.json file for all libs & apps
> * yes its possible to create a publishable lib (node or angular) which then generates a  seperate package.json
> * but there is no tooling support for those package.json files, they have to be maintained manuelly

Correct. But what would need to be maintained in those package.json's? I believe the libs should only define `peerDependency` ranges.

> * no integrated tooling support for semversioning

Correct. That actually surprised me :(
But I believe I solved that.

> * nx makes heavy use of its affected feature
>   
>   * this is especially helpful to drive down build times in CI environments by only building what was affected since the last PR completed.
>   * Making sure this works well with publishable libs reliably is not documented very well.
>     
>     * Assuming we have libs A, B, C + D
>     * B + C depends on A
>     * We now want to build  B + C + D:
>       `nx run-many --target build --with-deps --projects "B,C,D" --prod`
> * "-prod" flag when buidling

If A depends on B + C then you can just do `nx build C --with-deps` without `--with-deps` it'll complain and say you need to build B and C first. I found nx very helpful here.
`--prod` is only needed when bulding for release.

>   * when creating angular publishable libs in nx there is a production config generated to disable Ivy (NG9)

This is the currently recommended way to create ng9 libs, by the angular team

>   * when creating node libs there is no such config generated and using the --prod flag with those libraries will fail

I did not experience this. They simply used the normal config. But it's been a while and might have been through `run-many´

> Again, we are super happy with nx, but it is very apparent that the origins of nx is not in publishable packages.
> 
> Let me know if you have any questions, maybe we can assist.

Would your recommendation be to not use nx? Perhaps go with lerna instead? Perhaps just yarn 2?

I picked nx cause it seemed super simple to get the shared code feature working (which was the original problem I had that triggered this refactor 😄) and could scaffold things. But yes it does have its shortcomings

ntziolis:
@snebjorn first off. No this was not mean to discourage you from using nx, we are doing so as well also our publishable libs. Juts wanted to call out some surprising pitfalls we have found when using for publishable libs.

Note:
In regards to building I was talking about in CI pipelines. Since there you cannot react to some feedback from cli you need a build command that builds dependencies as needed automatically.

snebjorn:
fixes #397

Issue #421, opened 2020-05-09
Don't export the 'createInitialState' function from a store


theorlovsky:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

In case you have multiple states in one directory (which is not very good, but still), let's say it's a `state` directory with the following structure:
```
.
├── index.ts
├── state-a
│   ├── index.ts
│   ├── state-a.query.ts
│   ├── state-a.service.ts
│   └── state-a.store.ts
└── state-b
    ├── index.ts
    ├── state-b.query.ts
    ├── state-b.service.ts
    └── state-b.store.ts
```

And your `index.ts` files look like this (by default):
```
export * from './somewhere';
```

Then you're gonna have the following error in your top-level `index.ts`:
```
TS2308: Module './state-a' has already exported a member named 'createInitialState'.
Consider explicitly re-exporting to resolve the ambiguity.
```

## Expected behavior

I believe that in most cases you don't need the `createInitialState` function outside its state directory. If so, it might be a good idea to exclude this function from the corresponding `index.ts` file. Like this:
```
export * from './state-a.query';
export * from './state-a.service';
export { StateAState, StateAStore } from './state-a.store';
```

## What is the motivation / use case for changing the behavior?

I want the error above to be omitted by Akita, without the need to manage exports manually. What do you think about changing the schematics accordingly?

NetanelBasal:
Agree. There is no reason to export the `createInitialState` function. You are welcome to submit a PR. Thanks.

theorlovsky:
I'm not sure how to test `akita-cli`. Is there a way to build and use it locally?

NetanelBasal:
Run `node index.js`

theorlovsky:
Ok, thanks.

PR #423, opened 2020-05-10
improvement(akita-cli): stop exporting createInitialState from index.ts

theorlovsky:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #421 

- [ ] Yes
- [x] No

### DIFF
--- a/tools/akita-cli/templates/angular/index.tpl
+++ b/tools/akita-cli/templates/angular/index.tpl
@@ -1,5 +1,5 @@
 export * from './{{dashCase name}}.query';
-export * from './{{dashCase name}}.store';
+export { {{ pascalCase name }}State, {{pascalCase name}}Store } from './{{dashCase name}}.store';
 export * from './{{dashCase name}}.service';
 {{#switch storeType}}
 {{#case "Entity Store"}}
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 
 # dependencies
 /node_modules
+/tools/akita-cli/node_modules
 
 # IDEs and editors
 /.idea
--- a/tools/akita-cli/templates/ts/index.tpl
+++ b/tools/akita-cli/templates/ts/index.tpl
@@ -1,5 +1,5 @@
 export * from './{{dashCase name}}.query';
-export * from './{{dashCase name}}.store';
+export { {{ pascalCase name }}State, {{pascalCase name}}Store } from './{{dashCase name}}.store';
 export * from './{{dashCase name}}.service';
 {{#switch storeType}}
 {{#case "Entity Store"}}


theorlovsky:
@NetanelBasal please take a look at the both `yarn.lock` files.

The root one contained versions different from ones from the `package.json`. I ran `yarn install` and it changed `yarn.lock` accordingly. I'm not sure if those are valid changes.

The CLI's one contained errors. For example, a package called `get-_value`, which doesn't exist, it should be `get-value`. I'm not sure how this happened. Anyway, I renamed those invalid packages and ran `yarn install` from the CLI's directory.

Is it ok?

theorlovsky:
Also, I'm not sure if it's a breaking change. What do you think?

tools/akita-cli/templates/angular/index.tpl
```diff
@@ -1,8 +1,8 @@
 export * from './{{dashCase name}}.query';
-export * from './{{dashCase name}}.store';
+export { {{ pascalCase name }}State, {{pascalCase name}}Store } from './{{dashCase name}}.store';
```
NetanelBasal, line 2: Why not omit the export from `createInitialState`?

NetanelBasal:
> The root one contained versions different from ones from the package.json. I ran yarn install and it changed yarn.lock accordingly. I'm not sure if those are valid changes.

Which dependencies don't exist in the akita-cli folder?

NetanelBasal:
> Also, I'm not sure if changing this export is a breaking change. What do you think?

I don't think so. It'll not break an existing code.

NetanelBasal:
Note that you should also update the part of the schematic.

tools/akita-cli/templates/angular/index.tpl
theorlovsky: You might want to use this function in a service. So, it should be exported, but only inside a store. But if you find it a rare use case, I can change it.

NetanelBasal: Sounds ok.

theorlovsky:
> > The root one contained versions different from ones from the package.json. I ran yarn install and it changed yarn.lock accordingly. I'm not sure if those are valid changes.
> 
> Which dependencies don't exist in the akita-cli folder?

They all have `-_` in their names, while it should be just `-`: `get-_value`, `has-_value`, `set-_value`, `union-_value`, `unset-_value`. Yarn throws an error while trying to install those dependencies.

theorlovsky:
> Note that you should also update the part of the schematic.

Which schematic do you mean? I haven't found any that requires changes, but I might miss something.

NetanelBasal:
@snebjorn do you know something about the dependencies issues?

NetanelBasal:
There is https://github.com/datorama/akita/tree/master/libs/akita-schematics/src/ng-g but I see that only the CLI creates the `index.ts` file. Are you using the CLI?

theorlovsky:
> There is https://github.com/datorama/akita/tree/master/libs/akita-schematics/src/ng-g but I see that only the CLI creates the `index.ts` file. Are you using the CLI?

Yes, I am

snebjorn:
> @snebjorn do you know something about the dependencies issues?

@NetanelBasal I haven't gotten around to checking it, but I think it's related to https://github.com/datorama/akita/commit/31b1b879f8e965c0955a718c9af6c9b90bc16792

NetanelBasal:
Let us know when you can, please.

theorlovsky:
@NetanelBasal I can exclude those files from the PR, anyway they’re not directly related to the desired changes. After that I can open an issue for dependencies.

NetanelBasal:
Sure

theorlovsky:
Done

Issue #410, opened 2020-04-16
Improve typing of entity's queries


xbaun:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

When using `selecEntity` or `getEntity` (and other methods returning single entity objects) on an `EntityQuery`, the return value may be `undefined` if the Entity ID does not exist. 
The method signature of these functions does not reflect the possible `undefined` return value, so Typscript cannot correctly check/warn when a possibly undefined entity is returned and then accessed.

## Expected behavior

Methods that select/get a single entity on an EntityQuery should be extended to a union type with `undefined` as an additional return value.

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/typescript-wcf4a9

## What is the motivation / use case for changing the behavior?

Prevent the throwing of `Uncaught TypeError: Cannot read property 'XXX' of undefined` errors.

NetanelBasal:
I'm afraid this will lead to a breaking change.

xbaun:
I agree, but in the long run it's a small improvement that highly increases the fault tolerance.

NetanelBasal:
Agree, but it'll be added in v5.

NetanelBasal:
@xbaun it's a good time to add it.

xbaun:
I only refactored the return values of the query entity functions without compiler type checking support. For further releases two major tsconfig adjustments should be done:
- Enabling at least the compile option `strictNullChecks` (preferable would be to directly enable `strict`) and `noImplicitAny`.
- Update Typescript to 3.7 or above to use optional chaining for easier refactoring of all dependent tests and functions.
- Remove explicit `any` definitions as far as possible.

This will probably require more time to implement because a lot of code needs to be touched.

PR #439, opened 2020-06-02
fix(entity-store): improve type safety of entity store upsert operations

### LINKED ISSUE #436: Improve type safety for entity state operations
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

If `upsert` is invoked on an `EntityStore` for a previously non-exsiting entity, an empty entity (object) is passed to `UpdateStateCallback`. This behavior is correct under the assumption that all fields of the entity state model are declared optional (nullable). Non-optional fields are not taken into account and are thus implicitly initialized with `undefined` which is incorrect. Therefore, when a new entity is created by `upsert`, non-optional fields must be explicitly initialized by the caller in-place or through a separate callback.

## Expected behavior

Pass `undefined` as old state to [`UpdateStateCallback()`](https://github.com/datorama/akita/blob/master/libs/akita/src/lib/entityStore.ts#L219) or provide a separate entity initializing callback argument in `upsert`. If `undefined` is passed, it indicates that the caller has to initialize the entity state in-place. 

https://stackblitz.com/edit/angular-ivy-jzvwgs?file=src%2Fapp%2Fapp.component.ts

Improve/fix type safety for entity state operations: When using `upsert` in combination with https://github.com/datorama/akita/issues/153 a type safe entity store can be enforced.

xbaun:
Explicitly require an initialization callback if entities do not already exist.

Fix #436

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #436

- [x] Yes
- [ ] No

The function signature of `upsert` in `EntityStore` has changed. As a side effect, `runStoreAction` has been revised to support the new upsert function: The previous function overloading of `runStoreAction` caused several problems with the new `upsert` signature and was replaced by a discriminated action config object.


NetanelBasal:
We'll need to bump a major version to add this feature. There is no way to do it without introducing a breaking change?

xbaun:
I added backward compatibility to the upsert function but the `runStoreActions` are still a breaking change. The `runStoreActions` itself support the deprecated upsert call, but the action instantiation is still different to the original implementation.

NetanelBasal:
Can you summarize the changes, please?

xbaun:
A short summary:
* Refactored `upsert` function to safely update or insert entities by an optional initialization callback:
   - [new function signature](https://github.com/xbaun/akita/blob/master/libs/akita/src/lib/entityStore.ts#L213) with a `onCreate` callback for initialization if entity is previously not existing.
   - marked [old function signature](https://github.com/xbaun/akita/blob/master/libs/akita/src/lib/entityStore.ts#L223)  as deprecated because it violates type safety of the entity store.
* Rewrite [`libs/akita/src/lib/runStoreAction.ts`](https://github.com/xbaun/akita/blob/master/libs/akita/src/lib/runStoreAction.ts):
  - support type safe entity upsert actions.
  - refactored `upsertMany` into its own action: `StoreActions.UpsertManyEntities()`.
  - avoid type conflicts issued by the previous function overloading of `runStoreAction`.
* Adjusted docs and tests

NetanelBasal:
Two things:

- Do we have breaking changes?
- Quality Check / Build applications & libraries (pull_request) - is failing.

xbaun:
Yep, `runStoreAction` is the only breaking change. The new `upsert` function in `EntityStore` is fully backwards compatible.
The failed job is fixed: `ngc` seems not to like `this` references in a static context.

NetanelBasal:
> Yep, runStoreAction is the only breaking change

Can we prevent it?

xbaun:
Maybe, with a lot of conditional type generics in the `runStoreAction` definition (was my first attempt; and which make it really difficult to maintain), but I have not found any suitable solution. So I switched to a params object wich uses the action type as discriminant, so typescript itself can infer the appropriate type. It will therefore amount to a breaking change.

NetanelBasal:
If we bump to v5, I prefer to remove deprecations and keep the code clean. Please do it.

And please write in the https://github.com/datorama/akita/blob/master/BREAKING_CHANGES.md file the changes with before and after examples.

Thanks.

xbaun:
I updated BREAKING_CHANGES and removed deprecated functions.

I also rewrite `runStoreAction` a second time because there were still type check issues with the previous implementation. The idea know is to specify the store class, the operation to perform and execute it through a callback with the appropriate operator arguments. Please verify that the way `runStoreAction` works now is ok for you.

BREAKING_CHANGES.md
```diff
@@ -1,9 +1,47 @@
...
+// Before
+store.upsert([2, 3], (entity) => ({ isOpen: !entity.isOpen }));
+```
+
+```ts
+// After
+store.upsert(
+  [2, 3],
+  (entity) => ({ isOpen: !(entity?.isOpen ?? true) }),
+  (id, newState) => ({ id, ...newState, enabled: true })
```
NetanelBasal, line 18: Can we make it optional and use an empty object as before?

docs/docs/entities/entity-store.mdx
```diff
@@ -170,9 +170,13 @@ store.remove();
 Insert or update an entity. Creates a new entity when no entity matches the `id`; otherwise, it performs an update:
 
 ```ts
-store.upsert(id, { isOpen: true });
+store.upsert(id, { isOpen: true }, (id, newState) => ({ id, ...newState });
```
NetanelBasal, line 173: Regarding the previous comment, we can make this object as default: `(id, newState) => ({ id, ...newState });`

libs/akita/src/lib/entityStore.ts
```diff
@@ -204,28 +218,36 @@ export class EntityStore<S extends EntityState = any, EntityType = getEntityType
...
+    const predicate = (isUpdate) => (id) => hasEntity(this.entities, id) === isUpdate;
+    const baseClass = options.baseClass;
+    const isClassBased = isFunction(baseClass);
     const updateIds = toArray.filter(predicate(true));
-    const newEntities = toArray.filter(predicate(false)).map(id => {
-      let entity = isFunction(newState) ? newState({} as EntityType) : newState;
+    const newEntities = toArray.filter(predicate(false)).map((id) => {
+      const oldStateObj = this.entities[id as any];
+      const newStateObj = typeof newState === 'function' ? newState(oldStateObj) : newState;
+      const entity = onCreate(id, newStateObj ? newStateObj : undefined);
```
NetanelBasal, line 241: In your examples, it seems like you always have the `newState` param. But in reality, it can be `undefined`.

docs/docs/entities/entity-store.mdx
xbaun: This would only be possible if typescript knew in advance that all properties of a state are optional: But this is not possible. So it is mandatory to specify an initialization callback which returns the new state with all required properties.  At least it is an identity function that passes the previously updated state through unmodified if all non-optional fields are already set.

libs/akita/src/lib/entityStore.ts
xbaun: I missed that when I removed the backward compatibility. It can never be `undefined`, because the previous update callback always returns an object.

docs/docs/entities/entity-store.mdx
NetanelBasal: Don't you think it's annoying to do:
`store.upsert(id, { isOpen: true }, (id, newState) => ({ id, ...newState });` in each call?

libs/akita/src/lib/entityStore.ts
```diff
@@ -204,28 +218,36 @@ export class EntityStore<S extends EntityState = any, EntityType = getEntityType
...
-    const isClassBased = isFunction(options.baseClass);
+    const predicate = (isUpdate) => (id) => hasEntity(this.entities, id) === isUpdate;
+    const baseClass = options.baseClass;
+    const isClassBased = isFunction(baseClass);
     const updateIds = toArray.filter(predicate(true));
-    const newEntities = toArray.filter(predicate(false)).map(id => {
-      let entity = isFunction(newState) ? newState({} as EntityType) : newState;
+    const newEntities = toArray.filter(predicate(false)).map((id) => {
+      const oldStateObj = this.entities[id as any];
+      const newStateObj = typeof newState === 'function' ? newState(oldStateObj) : newState;
```
NetanelBasal, line 240: This line will break the existing code as `oldStateObj` could be `undefined` now.

NetanelBasal, line 240: - You pass the `newStateObj` to `onCreate` but it's not always the complete state as you can pass a partial update.
- I can't see the point of passing the `newStateObj` to `onCreate` as it will always be `undefined`.

xbaun: This is taken into account in the definition of the [`newState`](https://github.com/datorama/akita/blob/b487c5c42e1474e1ae5791e97f84fe7067f214b3/libs/akita/src/lib/types.ts#L37) callback type: The callback expects the old entity object or `undefined`. The problem with the current [tsconfig](https://github.com/datorama/akita/issues/410#issuecomment-640200763) is that typescript will not do a type check and throws no errors when accessing `undefined` values.

xbaun: The `newState` callback returns a partial update for the old state. If no old state exists the new state is passed in the `onCreate` callback to initialize the new entity, otherwise it is directly applied to the store.

But you indicated a mistake on my part: I forgot to check if 'onCreate' needs to be called.

xbaun: Ok, I think there is another error. I will rework the code.

xbaun: > * I can't see the point of passing the `newStateObj` to `onCreate` as it will always be `undefined`. ( in a case of a function )

I have ignored that all existing entities were filtered out before. Now it should be fine.

docs/docs/entities/entity-store.mdx
xbaun: I totally agree with you that it is annoying, but I wouldn't know how to write this as a default parameter and at the same time respecting non-optional state properties that require a custom initialization callback.

NetanelBasal: Why not make it optional and do something like:

```ts
if(!isFunction(onCreate)) {
  entity = { [this.idKey]: id }
}
```

xbaun: Because it would potentially produce incorrect entity states, if non optional properties are defined:
```ts
interface State {
   id: number;
   name: string;
}
```
Here, the `name` property would be `undefined` but according to its type definition it is never undefined. So typescript will not show a warning that the property is undefined and must be checked for `undefined` value before accessing it.

NetanelBasal: Yes, I understand it. But for some developers, it's doesn't matter. If someone wants this feature, they'll pass the onCreate function. Otherwise, it will work the same as it works now.

xbaun: Ok, that's a point, but then there should be a warning in the docs which implications it has to omit the `onCreate` callback.

Another possible compromise would be to create a derived entity store class in which all state properties are automatically enforced to be optional. Then `onCreate` could always be omitted in `upsert`.

NetanelBasal: Update the docs and mention it. Thanks.

NetanelBasal:
Please resolve the conflicts.

NetanelBasal:
I think that some of your changes cause the build to fail. Can you check it, please?

https://github.com/datorama/akita/runs/769598715?check_suite_focus=true

NetanelBasal:
I want to release the new version.

xbaun:
The compiler also throws this error with previous commits. It seems that Babel has a problem with node v12/v13, with node v10 its building without errors: https://github.com/nodejs/node/issues/32852#issuecomment-613655150

NetanelBasal:
Did you change the node version?

xbaun:
Nope, I don't have an explanation why it's failing now. But I have developed and tested locally against node v10.

NetanelBasal:
Try to downgrade to node 10, and test it.

xbaun:
I can also verify this error for previous commits (e.g. I checked akita@4.23.2: https://github.com/datorama/akita/commit/76e383f3712cc36af2f7936132c273e52836a80e, before mine).

NetanelBasal:
Thanks

simeyla:
This broke a tonne of my code with errors like:

     Property 'data' does not exist on type '{} | OrderSummaryReport'.

For example this is the pattern I was using:

     this.userSessionStore.upsert(id, state => {

                    if (state.savepoints)
                    {
                            // add new savepoint to list
                            return {
                                ...state,                                
                                savepoints: [...state.savepoints, newSavepointData]
                            };
                        }
                    }
                    else 
                    {
                        // create new session
                        return {
                            savepoints: [data]
                        };
                    }
                });

Now it complains that `savePoints` doesn't exist on `{} | UserSession`
Just wanted to mention that for me it was a breaking change. 

I can change it to `if ('savepoints' in state) { ... }` then it works again.

Or maybe I'm completely using the `upsert` method incorrectly!

xbaun:
@simeyla Sorry to hear that. This was introduced in v5 as a breaking change. This behaviour is intentional, because if the entity does not exist before, an empty object is passed. The problem is, if a state consists of nested properties, you have to check wether this object is empty (non existing) or already exists with all required properties properly initialized. E.g.:

```ts
interface UserSession {
  a: {
     b: number
  }
}
```
Then this would throw an error if the entity does not exist, because the state argument is `{}` and an undefined property `a` is accessed:
```ts
 this.userSessionStore.upsert(id, state => ({ a: { b: state.a.b++  }}))
```
So as you mentioned you have to use a type guard to check for the presence of the state.

simeyla:
@xbaun shame you can't just say `== {}`

xbaun:
Try this type guard:
```
function isUpdate<T>(a: any): a is T {
    return !(Object.keys(a).length === 0 && Object.getPrototypeOf(a).constructor === Object.prototype.constructor);
}
```

[Playground Link](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=2&pc=22#code/GYVwdgxgLglg9mABDAzgVQA4BMCGUCmAPACoB8AFDgFyI5gCeAlDTsiosYgN4BQi-iAE74oIQUgCE5APIAjAFb5oAOgDW+eikqNlAG3xgA5lAAWiALyXEABkQAyO4jmKVhkQAVBcKN-oZ80sDayhAIKFCCINBwghZWzkpQyhhePlB++CFhEVE+gowA3DwAvjw8MGAEgsA4EPiIAGJwcNx8AgAeNOGCFYYlZaFg4YjAzTRNLQA+AAVcxRbcyFg0AIwANIiwUPo0AORb+ruIxUXlwIjkqJi4BIQTFKNwjIytAoiDKHD6enCG5ABEIGweHw-w2j0YJUQ+F0KHqvDeHy+mV0vwBEGEILBI2akOKQA)

Issue #396, opened 2020-03-30
Log actions produces incorrect actions in redux devtools


danyo1399:
## I'm submitting a bug where logAction does not log actions accurately in redux dev tools

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior
When using logAction in certain scenarios, the actions dont correlate with the with the updates to state. Plus one action can be logged as two actions in redux devtools. See stackblitz below where you will see "set id" action logged twice for the one action

https://stackblitz.com/edit/angular-jfucsj

## Expected behavior
Actions should correlate with state changes + duplicate actions should not raised in dev tools for the one action

## Minimal reproduction of the problem with instructions
Open https://stackblitz.com/edit/angular-jfucsj 
have redux devtools installed.
open in new window so redux devtools binds to the app.
click the button.
then click the button again and see "set id" action logged twice for the one action

## What is the motivation / use case for changing the behavior?
The one thing i miss when moving from redux to Akita is the logging info you can get out of redux dev tools

## Environment

<pre><code>
Angular version: 8.2.14

Browser:
- [ X] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  stackblitz
- Platform:  mac / stackblitz

Others:

</code></pre>

chriszrc:
@danyo1399 Thanks for posting the stack blitz repro, I can definitely see that it's logging to actions. Have you been able to replicate the other behavior that you mention in this issue, namely that action names are not corresponding to the actions beginning executed?

PR #441, opened 2020-06-04
fix(devtools): fix multiple logging of same actions

xbaun:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A store update triggers query subscribers before releasing the current log action through `resetCustomAction()`:  So if a query subscriber itself triggers a `logAction()` after a previous one (which is still blocking/holding the log action), the previous will be logged again. This is caused by RxJS which is executing all subscribers synchronously and thus the `resetCustomAction()` is executed after `next()` has finished in `dispatch()` (and therefore after the query subscribers). 

Issue Number: #396

## What is the new behavior?

The store `BehaviourSubject` itself holds beside the current state the associated log action and sends it to the Redux DevTools before query subscribers are called.

- [ ] Yes
- [X] No

### DIFF
--- a/libs/akita/src/lib/dispatchers.ts
+++ b/libs/akita/src/lib/dispatchers.ts
@@ -1,11 +1,12 @@
 import { ReplaySubject, Subject } from 'rxjs';
+import { StoreSnapshotAction } from './store';
 
 // @internal
 export const $$deleteStore = new Subject<string>();
 // @internal
 export const $$addStore = new ReplaySubject<string>(50, 5000);
 // @internal
-export const $$updateStore = new Subject<string>();
+export const $$updateStore = new Subject<{ storeName: string; action: StoreSnapshotAction }>();
 
 // @internal
 export function dispatchDeleted(storeName: string) {
@@ -18,6 +19,6 @@ export function dispatchAdded(storeName: string) {
 }
 
 // @internal
-export function dispatchUpdate(storeName: string) {
-  $$updateStore.next(storeName);
+export function dispatchUpdate(storeName: string, action: StoreSnapshotAction) {
+  $$updateStore.next({ storeName, action });
 }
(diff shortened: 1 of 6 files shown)


libs/akita/src/lib/store.ts
```diff
@@ -176,7 +184,12 @@ export class Store<S = any> {
     }
 
     if (!this.store) {
-      this.store = new BehaviorSubject(this.storeValue);
+      this.store = new BehaviorSubject({ state: this.storeValue });
+      this.store.subscribe(({ action }) => {
```
NetanelBasal, line 188: We need to run it only in dev-mode.

libs/akita/src/lib/dispatchers.ts
```diff
@@ -18,6 +19,6 @@ export function dispatchAdded(storeName: string) {
 }
 
 // @internal
-export function dispatchUpdate(storeName: string) {
-  $$updateStore.next(storeName);
+export function dispatchUpdate(storeName: string, action: StoreSnapshotAction) {
```
NetanelBasal, line 22: Thanks for your PR.

In those days, I didn't have time to write tests for the devtools feature. Do you think this is something you can add?

You can mock the dev-tools and test that they get the correct actions.

xbaun: Added a test suite for devtools with some basic log action tests.

libs/akita/src/lib/store.ts
xbaun: Will now check if `isDev()` is true and only then subscribe to the store to dispatch logs.

libs/akita/src/lib/dispatchers.ts
NetanelBasal: Thanks!

We'll go out with v5 next week. If you think about more things that can be improved and require breaking changes, let me know.

PR #451, opened 2020-06-17
fix(run-store-action): add untyped store support

### LINKED ISSUE #449: Benefit of using RunStoreAction vs injecting the service and calling the method directly?
## I'm submitting a...

<pre><code>
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[x] Support request
[ ] Other... Please describe:
</code></pre>

Assuming I have a data payload coming from signalR, that calls this method:
```
    private applySignalrPayload(store, entityId, payload): void {

        switch (store) {
            case AkitaEntityStores.DeviceTelemetry:
                this.deviceTelemetryService.upsert(entityId, payload);
                break;
            case AkitaEntityStores.DeviceAlerts:
                this.deviceAlertService.upsert(entityId, payload);
                break;
            default:
                this.logger.warn('No matching Akita Entity Store was found to update.');
        }
}
```

Alternatively, how is that better or worse than this:
```
    private applySignalrPayload(store, entityId, payload): void {

        runStoreAction(store, StoreActions.UpsertEntities, {
            payload: {
                data:      payload,
                entityIds: entityId
            }
        });
    }
```

The only different is that using   `runStoreAction` requires putting the action i.e. `upsert` on the store and calling `super.upsert`. Whereas injecting the service requires putting the action on the service.

What are the pros/cons of using `runStoreAction` vs injecting the service and updating this way? They accomplish the same thing so I'm wondering what the use case of `runStoreAction` is when we can inject the services instead.

xbaun:
fix #449

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/docs/docs/additional/events.mdx
+++ b/docs/docs/additional/events.mdx
@@ -4,7 +4,7 @@ title: Event-based APIs
 
 One of the recurring requests we got, was to simplify and improve the experience of working with event-based APIs such as web-sockets.
 
-To make it easier for you, we’ve added a new API method — `runStoreAction`:
+To make it easier for you, we’ve added a new API method — `runStoreAction` and `runEntityStoreAction`:
 
 ```ts
 import { runStoreAction, StoreActions, runEntityStoreAction, EntityStoreActions  } from '@datorama/akita';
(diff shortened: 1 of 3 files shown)


xbaun:
Now `runStoreAction` can be invoked with only the store name, but with type checking disabled (the store state is `any`):

```ts
runStoreAction('books', StoreAction.Update, (update) => update({ filter: 'COMPLETE' }));
```

@NetanelBasal Can you verify that this is the desired behaviour, please? Maybe we should add some more tests.

NetanelBasal:
LGTM, we just need to update the docs.

@hosneyMalkawi, please take a look to see if something is missing. Thanks.

hakimio:
- Might be good to have tests for `AddEntities` and `RemoveEntities` actions when store is specified as string.
- Since the store parameter is no longer of type class,  might be a good idea to rename it from `storeClass` to just `store`.

xbaun:
Added more tests to cover all possible `runStoreAction` invocations by store class/name and updated docs.

silverstory:
My akita app is not working anymore when upgraded to angular version 10
and also upgraded akita to
@datorama/akita version 5.2.5

Help me! Now!!!

App error is complaining about line 4 and line 8 in this file:
https://github.com/silverstory/akita-socketio/blob/master/src/app/livefeed-list.service.ts

Link to app
https://github.com/silverstory/akita-socketio

PR #476, opened 2020-07-01
feat(entity-store): track entity id updates

xbaun:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?

Add rxjs operator to track entity id updates and automatically re-evaluate the query.

- [ ] Yes
- [x] No

### DIFF
--- a/libs/akita/src/__tests__/queryEntity.spec.ts
+++ b/libs/akita/src/__tests__/queryEntity.spec.ts
@@ -3,6 +3,7 @@ import { QueryConfig, SortBy } from '../lib/queryConfig';
 import { QueryEntity } from '../lib/queryEntity';
 import { Order } from '../lib/sort';
 import { isObject } from '../lib/isObject';
+import { trackIdChanges } from '../lib/trackIdChanges';
 import { cot, createTodos, ct, Todo, TodosStore } from './setup';
 import { getInitialEntitiesState } from '..';
 
@@ -30,19 +31,19 @@ describe('Entities Query', () => {
 
   describe('Select', () => {
     it('should select ids', () => {
-      sub = query.select(state => state.ids).subscribe(spy);
+      sub = query.select((state) => state.ids).subscribe(spy);
       expect(spy).toHaveBeenCalledTimes(1);
       expect(spy).toHaveBeenCalledWith(store._value().ids);
     });
 
     it('should select entities', () => {
-      sub = query.select(state => state.entities).subscribe(spy);
+      sub = query.select((state) => state.entities).subscribe(spy);
       expect(spy).toHaveBeenCalledTimes(1);
       expect(spy).toHaveBeenCalledWith(store._value().entities);
     });
 
     it('should fire only when updating the state', () => {
-      sub = query.select(state => state.entities).subscribe(spy);
+      sub = query.select((state) => state.entities).subscribe(spy);
       let todo = cot();
       store.add(todo);
       expect(store._value().entities[1]).toBe(todo);
@@ -107,7 +108,7 @@ describe('Entities Query', () => {
     it('should select slice from entity', () => {
       let todo = cot();
       store.add(todo);
-      sub = query.selectEntity(1, entity => entity.title).subscribe(spy);
+      sub = query.selectEntity(1, (entity) => entity.title).subscribe(spy);
       expect(spy).toHaveBeenCalledTimes(1);
       expect(ga(spy)).toBe(todo.title);
     });
@@ -115,7 +116,7 @@ describe('Entities Query', () => {
     it('should not fire when the selected value does not changed', () => {
       let todo = cot();
       store.add(todo);
-      sub = query.selectEntity(1, entity => entity.title).subscribe(spy);
+      sub = query.selectEntity(1, (entity) => entity.title).subscribe(spy);
       store.update(1, { completed: true });
       expect(spy).toHaveBeenCalledTimes(1);
     });
@@ -123,15 +124,15 @@ describe('Entities Query', () => {
     it('should fire when the selected value changed', () => {
       let todo = cot();
       store.add(todo);
-      sub = query.selectEntity(1, entity => entity.title).subscribe(spy);
+      sub = query.selectEntity(1, (entity) => entity.title).subscribe(spy);
       store.update(1, { title: 'changed' });
       expect(spy).toHaveBeenCalledTimes(2);
     });
 
     it('should not throw when the entity does not exists', () => {
       sub = query
-        .selectEntity(2, entity => entity.title)
-        .subscribe(entity => {
+        .selectEntity(2, (entity) => entity.title)
+        .subscribe((entity) => {
           expect(entity).toBe(undefined);
         });
     });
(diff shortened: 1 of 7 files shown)


libs/akita/src/lib/trackIdChanges.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export function trackIdChanges<K extends EntityStore<S, T>, S extends EntityState<T> = getEntityStoreState<K>, T = getEntityType<S>>(store: K): MonoTypeOperatorFunction<T> {
+  return (source) => source.lift<T>(new TrackIdChanges(store));
+}
+
+class TrackIdChanges<K extends EntityStore<S, T>, S extends EntityState<T>, T = getEntityType<S>> implements Operator<T, T> {
+  readonly query: QueryEntity<S, T>;
+
+  constructor(readonly store: K) {
+    this.query = new QueryEntity((store as unknown) as EntityStore<S>);
```
NetanelBasal, line 26: Why create new query entity?

libs/akita/src/lib/trackIdChanges.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @example
+ *
+ *   query.selectEntity(1).pipe(trackIdChanges(store)).subscribe(entity => { ... })
+ *
+ */
+export function trackIdChanges<K extends EntityStore<S, T>, S extends EntityState<T> = getEntityStoreState<K>, T = getEntityType<S>>(store: K): MonoTypeOperatorFunction<T> {
+  return (source) => source.lift<T>(new TrackIdChanges(store));
+}
+
+class TrackIdChanges<K extends EntityStore<S, T>, S extends EntityState<T>, T = getEntityType<S>> implements Operator<T, T> {
```
NetanelBasal, line 22: Why do we need to pass the store? We can keep it simple:

```ts
entity$ = this.query.selectEntity(id).pipe(trackIdChanges(id))
```

libs/akita/src/lib/trackIdChanges.ts
xbaun: So I can use `selectEntity` for the re-evaluation of the query. I wanted to avoid writing redundant code.

libs/akita/src/lib/entityStore.ts
```diff
@@ -204,9 +218,17 @@ export class EntityStore<S extends EntityState = any, EntityType = getEntityType
...
         newStateOrFn,
         producerFn: this._producerFn,
+        onEntityIdChanges: (oldId: IDType, newId: IDType) => {
+          entityIdChanged = { oldId, newId };
+          this.entityIdChanges.next({ ...entityIdChanged, pending: true });
+        },
       })
     );
 
+    if (entityIdChanged) {
```
NetanelBasal, line 228: Why not aggregate the ids and call it once here?

libs/akita/src/lib/trackIdChanges.ts
xbaun: Because the operator needs to subscribe to the `selectEntityIdChanges$` observable.

libs/akita/src/lib/entityStore.ts
xbaun: Because undefined values would be emitted by the query when the id is updated: In fact the old entity is delete and thus undefined is emitted by the previous query before the new one emit the updated entity. This means the tracking query is notified about the id change (`pending=true`), so that undefined values can be suppressed and only then the updated entity is emitted again.

libs/akita/src/lib/trackIdChanges.ts
NetanelBasal: So pass the query, why do you need the store?

libs/akita/src/lib/trackIdChanges.ts
NetanelBasal: See previous comment

libs/akita/src/lib/trackIdChanges.ts
xbaun: It's no problem to use the query instead. I change it.

libs/akita/src/lib/trackIdChanges.ts
```diff
@@ -0,0 +1,57 @@
+import { EntityState, EntityStore, getEntityType, getIDType, ID, QueryEntity } from '@datorama/akita';
+import { merge, MonoTypeOperatorFunction, Observable, of, Operator, Subscriber, TeardownLogic, UnaryFunction } from 'rxjs';
+import { distinctUntilChanged, filter, first, switchMap, tap } from 'rxjs/operators';
+import { OperatorFunction } from 'rxjs/src/internal/types';
+
+type getEntityStoreState<T extends EntityStore<any>> = T extends EntityStore<infer S> ? S : never;
+
+/**
+ * Track id updates of an entity and re-evaluation the query with the changed entity id.
+ * Hint: Don't place the operator after other operators in the same pipeline as those will be skipped on
```
NetanelBasal, line 10: Also, it can be used with the selection operator. `(selectEntity(1, e => e.title))`

libs/akita/src/lib/trackIdChanges.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.query = new QueryEntity((store as unknown) as EntityStore<S>);
+  }
+
+  call(subscriber: Subscriber<T>, source: Observable<T>): TeardownLogic {
+    return source
+      .pipe(
+        first(),
+        switchMap((entity) => {
+          let currId = entity[this.store.config.idKey];
+          let select = this.query.selectEntity(currId);
```
NetanelBasal, line 35: Where do you use the `select` variable? The only place I see is overriding it.

xbaun: I missed to remove it. The variable become redundant after using the `pending` state.

libs/akita/src/__tests__/queryEntity.spec.ts
```diff
@@ -1098,13 +1099,46 @@ describe('selectAll - limit to and filterBy and sorting', () => {
...
+    expect(Object.keys(store._value().entities).length).toBe(1);
+    expect(Object.keys(store._value().ids).length).toBe(1);
+
+    store.update(1, { id: 2, title: 'a' });
+    expect(res.id).toBe(2);
+    expect(res.title).toBe('a');
+    expect(Object.keys(store._value().entities).length).toBe(1);
+    expect(Object.keys(store._value().ids).length).toBe(1);
+
+    store.update(2, { id: 3, title: 'a' });
```
NetanelBasal, line 1138: Please, add one section that updates one of the current entity props ( not the `id`)

NetanelBasal:
Let's also add it to the QueryConfig as optional. Most of the users will not need it so we can make the functionality lazy. 

`trackEntityIdChanges: boolean`

xbaun:
If no one uses this operator, then there are no subscribers to the `entityIdChanges$` observable and thus is pretty lazy without explicitly disabling functionality through the config.

NetanelBasal:
You're right; it'll anyway only run if someone changes the id.

libs/akita/src/lib/trackIdChanges.ts
```diff
@@ -0,0 +1,52 @@
+import { EntityState, EntityStore, getEntityType, getIDType, ID, QueryEntity } from '@datorama/akita';
```
NetanelBasal, line 1: Expose it in public_api

xbaun: I added the export. There was also an import error by using '@datorama/akita' instead of relative paths.

xbaun:
Added a first draft for the docs.

PR #519, opened 2020-08-26
Add disableTracking to StoreConfig

hrvbernardic:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently there is no way to exclude certain store from devtools tracking through store config.

Issue Number: N/A

## What is the new behavior?

With new disableTracking flag it is possible to disable tracking for a certain store allowing you to use it in a library and having an option to not pollute applications devtools tracking or providing an option to turn it on when needed.

- [ ] Yes
- [x ] No

### DIFF
--- a/libs/akita/src/__tests__/setup.ts
+++ b/libs/akita/src/__tests__/setup.ts
@@ -109,3 +109,6 @@ export function createWidget(id) {
 }
 
 export const tick = () => new Promise(process.nextTick);
+
+@StoreConfig({ name: 'disabledTrackingConfig', disableTracking: true })
+export class DisabledTrackingConfigStore extends EntityStore<State> {}
--- a/docs/docs/config.mdx
+++ b/docs/docs/config.mdx
@@ -62,6 +62,10 @@ See [caching support](../additional/cache) section.
 ### `idKey`
 A custom `idKey` for `EntityStore` - see [EntityId](../entities/entity-store#entity-id) section.
 
+### `disableTracking`
+Option to disable devtools tracking.
+
+
 :::info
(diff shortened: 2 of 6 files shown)


libs/akita/src/__tests__/setup.ts
```diff
@@ -109,3 +109,6 @@ export function createWidget(id) {
 }
 
 export const tick = () => new Promise(process.nextTick);
+
+@StoreConfig({ name: 'disabledTrackingConfig', disableTracking: true })
```
NetanelBasal, line 113: I prefer not using devtools configuration in the store. It should be in the devtools options.[ ](https://datorama.github.io/akita/docs/enhancers/devtools/#options)

hrvbernardic: Ok, so my idea was to enable using Akita in libraries for complex components. This way I could disable it by default and the application developer wouldn't see them unless he chose to through let's say component's module configuration. I suppose you want me to add storeBlacklist to devtools options? In that case i should do it by providing storeBlacklistPattern since i will be relying on storeName's prefix.

NetanelBasal: Add storeBlacklist and support both strings and cb function.

NetanelBasal: And add yourself to the contributors' list

hrvbernardic: Ok, deal.

hrvbernardic: I'll create a new PR

Issue #562, opened 2020-11-25
[Question] Concatenate extra URL segments to the base URL in the NgEntityService


juanmesa2097:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[x] Support request
[ ] Other... Please describe:
</code></pre>

## Problem

I'm currently using the NgEntityService and I need to consume an API that have endpoints like this:

**PUT** `example.com/api/v1/usuarios/1/detalles`
**PATCH** `example.com/api/v1/usuarios/1/status`
**GET** `example.com/api/v1/municipios/departamento/1`

## Example

1. Set the base URL

app.module

```ts
providers: [
  ...
  {
      provide: NG_ENTITY_SERVICE_CONFIG,
      useValue: {
        baseUrl: environment.apiUrl,
      },
  }
]
```

2. Add resource name

users.service

```ts
@NgEntityServiceConfig({
  resourceName: 'usuarios',
})
@Injectable({ providedIn: 'root' })
export class UsersService extends NgEntityService<UsersState> {
  constructor(protected store: UsersStore) {
    super(store);
  }
}
```

3. Call service

account.page.ts

```ts
updateUserDetails(): void {
  this.usersService
    .put<User>({      
      extras: ['detalles'],      // <---   Is there a way to do this?
      mapResponseFn: (res) => res.data as User,
    })
    .subscribe();
}
```

#### WORKAROUND

```ts
updateUserDetails(): void {
  this.usersService
    .put<User>({      
      url: `${environment.apiUrl}/usuarios/detalles`, 
      mapResponseFn: (res) => res.data as User,
    })
    .subscribe();
}
```

## Question

How can I concatenate extra endpoints to the API resource?

## Environment

```
Angular 11.0.0
```

P.D. I know I could use query parameters for some of the examples but I don't have control over the API.

Thanks in advance!

NetanelBasal:
There is no way to do it. You're welcome to submit a PR.

juanmesa2097:
> There is no way to do it. You're welcome to submit a PR.

Ok, thank you. I'll take some time this weekend to work on it!

PR #564, opened 2020-11-27
Feature/entity service sub resources

juanmesa2097:
## PR Checklist

- [x] Tests for the changes have been added (for bug fixes/features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

**Some possible use cases**
```
/posts/:postId/comments/:commentId
/users/:userId/articles/:articleId

/users/:userName/repos
/repos/:repoName/users
```
Currently, there's no way to specify sub-resources in the NgEntityService, so if you need a use case like the ones above, you'll have to override the whole API URL in the `url` property.

Issue Number: #562

## What is the new behavior?

With the new property "subResources" in the `httpConfig` type, now is possible to specify sub-resources in a form of an array. For example, if you need to get the articles of a certain user, you can do as follows:

```ts
// base: "https://api.example.com/v1"
// main resource: "users"

const userId = 1;
const articleId = 99;

this.usersService.get(userId, {
  urlPostfix: `articles/${articleId.toString}`
}).subscribe();
```

the final URL is: `https://api.example.com/v1/users/1/articles/99`

`users/1`   ---> main resource
`articles/99 `  ---> sub resource

- [ ] Yes
- [x] No

### DIFF
--- a/libs/akita-ng-entity-service/src/lib/types.ts
+++ b/libs/akita-ng-entity-service/src/lib/types.ts
@@ -28,7 +28,7 @@ export type HttpConfig<Entity = any> = {
   params?: _HttpParams;
   headers?: _HttpHeaders;
   url?: string;
-  subResources?: string[];
+  urlPostfix?: string;
   mapResponseFn?: (res: any) => Entity | Entity[];
 };
 
--- a/libs/akita-ng-entity-service/src/lib/types.ts
+++ b/libs/akita-ng-entity-service/src/lib/types.ts
@@ -28,6 +28,7 @@ export type HttpConfig<Entity = any> = {
   params?: _HttpParams;
   headers?: _HttpHeaders;
   url?: string;
+  subResources?: string[];
   mapResponseFn?: (res: any) => Entity | Entity[];
 };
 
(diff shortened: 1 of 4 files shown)


libs/akita-ng-entity-service/src/lib/types.ts
```diff
@@ -28,6 +28,7 @@ export type HttpConfig<Entity = any> = {
   params?: _HttpParams;
   headers?: _HttpHeaders;
   url?: string;
+  subResources?: string[];
```
NetanelBasal, line 31: Why use an array, and why couple it to "resources". I think we can change it to:

`urlPostfix: string;`

`urlPostfix: '/some/3/page/3'`

libs/akita-ng-entity-service/src/lib/types.ts
```diff
@@ -28,6 +28,7 @@ export type HttpConfig<Entity = any> = {
   params?: _HttpParams;
```
NetanelBasal, line 28: Please update the docs

libs/akita-ng-entity-service/src/lib/types.ts
juanmesa2097: I agree. I did it in a form of an array because I had in my mind `router.navigate([])` but concatenating the parameters to the string is just as good and "urlPostfix" definitely sounds better. I'll change it.

libs/akita-ng-entity-service/src/lib/types.ts
juanmesa2097: I missed it. I'm on it

NetanelBasal (approved):

Issue #571, opened 2020-12-04
Create effects package for Akita


NetanelBasal:
Developers that comes from ngrx sometimes want similar functionality as effects in ngrx. I think it'll be nice to develop a package that provides this functionality if someone wants to use it.

Here's what I imagine: ( I'm open to any other suggestions )

```ts
import { Actions } from '@datorama/akita-effects';

export class TodosComponent {
  todos$ = this.query.selectAll();

  constructor(private actions: Actions, private query: TodosQuery) {}

  ngOnInit() {
    this.actions.dispatch(fetchTodos());
  }
}

import { Actions, createEffect } from '@datorama/akita-effects';

class TodosEffects {
  constructor(private actions$: Actions, 
              private service: TodosService) {}

  fetchTodos$ = createEffect(() =>
    this.actions$.pipe(
      ofType(fetchTodos),
      switchMap(() => this.service.fetchTodos())
    )
  );
}

class TodosService {
  fetchTodos() {
    return this.http.get("todos").pipe(
      tap(todos => this.store.set(todos))
    );
  }
}
```

You can save some work and use this package for the actions:

https://github.com/cartant/ts-action/blob/master/packages/ts-action-operators/README.md
https://github.com/cartant/ts-action/blob/master/packages/ts-action/README.md

hakimio:
How is `this.actions.dispatch(fetchTodos())` better than just calling `this.todosService.fetchTodos()` directly? Can you explain what is the problem `effects` solve here? Seems like useless boilerplate to me.

Also, where do you define your `fetchTodos()`/`fetchTodos` in this example? And should it be ` this.actions.dispatch(fetchTodos())` or ` this.actions.dispatch(fetchTodos);`?

NetanelBasal:
I didn't say that it's better :)

Some of the developers coming from ngrx prefer this pattern. You define the `actions` in the `actions` file. 

Imagine a case where you want to do some task when the user logs out. You listen to `authQuery.isLoggedIn$`, but where is the best place to put the subscription? That's where an effect can be useful.

hakimio:
> Imagine a case where you want to do some task when the user logs out. You listen to `authQuery.isLoggedIn$`, but where is the best place to put the subscription? That's where an effect can be useful.

Depends on the task. If it's UI update task, most likely in the UI component, if it's some resource clean-up task, than in the service responsible for managing those resources.

Can you illustrate with some code, how actions are helpful in this case?

NetanelBasal:
[...]

PR #580, opened 2020-12-18
feat(akita-ng-effects): add library akita-ng-effects

yackinn:
initial setup## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

https://github.com/datorama/akita/issues/571

Issue Number: #571

## What is the new behavior?

Effects are included. Listen and act upon action events with rxjs operators.

- [ ] Yes
- [x] No

## Other information
The compiler is showing two errors depending on the Angular version and compiler settings.

This store.ts probably needs an injectable decorator in ivy Angular versions.

Error: libs/akita/src/lib/store.ts:48:14 - error NG2007: Class is using Angular features but is not decorated. Please add an explicit Angular decorator.
    
    48 export class Store<S = any> {

In some circumstances I saw the error that no function calls should be used inside a decorator.
AkitaNgEffects.forRoot([]) shouldn't be any different than RouterModule.forRoot() if you compare them as function calls.


apps/angular-store-app/src/app/app.module.ts
```diff
@@ -27,7 +29,11 @@ import { AkitaNgDevtools } from '@datorama/akita-ngdevtools';
     CartModule,
     AuthModule,
     AkitaNgRouterStoreModule,
-    AkitaNgDevtools.forRoot()
+    AkitaNgDevtools.forRoot(),
+    // ERROR in Error during template compile of 'AppModule'
```
NetanelBasal, line 33: What do u mean by "An Angular update will fix this"?

apps/angular-store-app/src/app/cart/state/cart.effects.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Actions, Effect, ofType } from '@datorama/akita-ng-effects';
+import { CartActions } from './cart.actions';
+import { map, switchMap, tap } from 'rxjs/operators';
+import { CartStore } from './cart.store';
+import { of, timer } from 'rxjs';
+
+@Injectable()
+export class CartEffects {
+  constructor(private actions$: Actions, private cartStore: CartStore) {
+    console.log(this);
```
NetanelBasal, line 11: remove console.log, please

apps/angular-store-app/src/app/cart/state/cart.effects.ts
```diff
@@ -0,0 +1,32 @@
...
+    ofType(CartActions.removeItem),
+    tap((val) => console.log('[Cart Effects] removeItem$', val)),
+    switchMap((test) => {
+      // do async operations
+      this.cartStore.remove(test.productId);
+      return of(this.actions$.dispatch(CartActions.removeItemSuccess()));
+    })
+  );
+
+  @Effect()
```
NetanelBasal, line 25: Use the `createEffect` function in one of these methods, please.

apps/angular-store-app/src/app/cart/state/cart.effects.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  @Effect()
+  removeItem$ = this.actions$.pipe(
+    ofType(CartActions.removeItem),
+    tap((val) => console.log('[Cart Effects] removeItem$', val)),
+    switchMap((test) => {
+      // do async operations
+      this.cartStore.remove(test.productId);
+      return of(this.actions$.dispatch(CartActions.removeItemSuccess()));
```
NetanelBasal, line 21: I think it'll be better to follow the `ngrx/effects` style that does the following:
- By default, it expects the `effect` to return a new action and dispatches it internally
- You can pass `{ dispatch: false }` as the second argument to cancel the previous behavior.

libs/akita-ng-effects/README.md
```diff
@@ -0,0 +1,98 @@
+# Akita Effects
```
NetanelBasal, line 1: Add it also to our official docs, please. You can find it in the `docs` folder.

libs/akita-ng-effects/README.md
```diff
@@ -0,0 +1,98 @@
...
+    private actions$: Actions, 
+    private navigationService: NavigationService, 
+    private store: NavigationStore
+  ) {}
+
+  // can also be immplemented with createEffect() function
+  @Effect()
+  loadMainNavigation$ = this.actions$.pipe(
+    ofType(LOAD_MAIN_NAVIGATION),
+    switchMap(_ => this.navigationService.LOAD_MAIN_NAVIGATION().pipe(
```
NetanelBasal, line 64:
```suggestion
    switchMap(_ => this.navigationService.loadMainNavigation().pipe(
```

libs/akita-ng-effects/README.md
```diff
@@ -0,0 +1,98 @@
...
+      tap(mainNav => this.actions$.dispatch(LOAD_MAIN_NAVIGATION_SUCCESS({ mainNav }))))
```
NetanelBasal, line 65:
```suggestion
      tap(mainNav => this.actions$.dispatch(loadMainNavigationSuccess({ mainNav }))))
```

libs/akita-ng-effects/README.md
```diff
@@ -0,0 +1,98 @@
...
+    )
+  );
+
+  @Effect()
+  loadMainNavigationSuccess$ = this.actions$.pipe(
+    ofType(LOAD_MAIN_NAVIGATION_SUCCESS),
```
NetanelBasal, line 71:
```suggestion
    ofType(loadMainNavigationSuccess),
```

libs/akita-ng-effects/README.md
```diff
@@ -0,0 +1,98 @@
...
+export class InitRouterGuard implements CanActivate {
+  constructor(
+    private navigationQuery: NavigationQuery, 
+    private actions: Actions
+  ) {}
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean | UrlTree> {
+    return this.navigationQuery.isNavInitialized$.pipe(
+      tap((isInit) => {
+        if (!isInit) this.actions.dispatch(LOAD_MAIN_NAVIGATION());
```
NetanelBasal, line 91:
```suggestion
        if (!isInit) this.actions.dispatch(loadMainNavigation());
```

libs/akita-ng-effects/src/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { Actions } from '@datorama/akita-ng-effects';
+import { fakeAsync, flush } from '@angular/core/testing';
+
+describe('Actions', () => {
+  it('should notify on dispatched event', fakeAsync(() => {
+    const actions = new Actions();
```
NetanelBasal, line 6: Why we need `fakeAsync` here? It's sync operation.

libs/akita-ng-effects/src/__tests__/create-effect.util.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  it('should create an effect with effect metadata', () => {
+    const effect = createEffect(() => of('sample'));
+
+    expect(effect).toMatchObject({ isEffect: true });
+    expect(effect).toHaveProperty('name');
+  });
+
+  it('should emit a value upon subscription', (done) => {
+    createEffect(() =>
+      of('sample').pipe(
```
NetanelBasal, line 15: Why we need `done` here? It's sync operation.

libs/akita-ng-effects/src/__tests__/effect.decorator.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { Effect } from '@datorama/akita-ng-effects';
+import { of } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+describe('@Effect Decorator', () => {
+  it('should create ab effect with metadata', (done) => {
+    class TestClass {
+      @Effect()
+      testEffect$ = of('sample')
```
NetanelBasal, line 9: Why we need `done` here? It's sync operation.

libs/akita-ng-effects/src/lib/akita-ng-effects.module.ts
```diff
@@ -0,0 +1,60 @@
...
+          useFactory: createEffects,
+          deps: [Injector, _FEATURE_EFFECTS]
+        }
+      ]
+    };
+  }
+}
+
+export function createEffects(injector: Injector, effects: Type<any>[]): any[] {
+  const mergedEffects: Type<any>[] = [];
```
NetanelBasal, line 50: Why we need `mergedEffects`? We can just call `createEffectInstances(injector, effects)`

NetanelBasal (changes requested):
Great work!

See my comments.

One more thing. Do you think we can add support for registering effects at the component providers level? It'll be useful to free some memory when the component is destroyed, and we don't need to listen anymore.

apps/angular-store-app/src/app/app.module.ts
yackinn: I tried to run the project with multiple Angular versions for the akita store app.
Angular 8 had thrown an error.
Ivy versions worked fine without issues.
I haven't seen this error before.
Why wouldn't it be possible to use a function call within a decorator?

I'm not sure how to deal with this.
We can update the projects Angular version. That would also take a few changes to the code base.
I tried that already.

NetanelBasal: Ok, I will check it when you finish

libs/akita-ng-effects/src/__tests__/actions.spec.ts
yackinn: As far as I'm concerned this is async.
The subject publishes the event and the assertion should run inside the async callback of the subscription.
It results in a false positive once the fakeAsync is removed.
The assertion won't run.

libs/akita-ng-effects/src/__tests__/create-effect.util.spec.ts
yackinn: As far as I'm concerned this is async.
The test will finish before the assertion has run.
It results in a false positive once the fakeAsync is removed.

libs/akita-ng-effects/src/__tests__/effect.decorator.spec.ts
yackinn: As far as I'm concerned this is async.
The test will finish before the assertion inside the tap operator has run.
It results in a false positive once the fakeAsync is removed.
In this test the assertion inside the tap operator isn't crucial I guess.

libs/akita-ng-effects/src/lib/akita-ng-effects.module.ts
yackinn: Absolutely correct. That was a leftover from an earlier different approach.

apps/angular-store-app/src/app/cart/state/cart.effects.ts
yackinn: removed

apps/angular-store-app/src/app/cart/state/cart.effects.ts
yackinn: Good call.
When this pull request is merged I might have a few more feature adaptions for the effects.
We can chat later about this.

apps/angular-store-app/src/app/cart/state/cart.effects.ts
yackinn: Changed.

libs/akita-ng-effects/README.md
yackinn: It's included now.

yackinn:
Yes we can add a feature to provide the effects on a component level. Let me check.

Do you have any suggestions for the implementation?

If the effect isn't provided in any module we need to make sure to unsubscribe from the effects when the component is destroyed.
Can we use a decorator for such an effect?

@EffectsCollection()
class TestEffects {
}

libs/akita-ng-effects/README.md
yackinn: changed

libs/akita-ng-effects/src/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { Actions } from '@datorama/akita-ng-effects';
+import { fakeAsync, flush } from '@angular/core/testing';
+import { Subject } from 'rxjs';
+
+describe('Actions', () => {
```
NetanelBasal, line 5: https://stackblitz.com/edit/angular-testing-qknhej?file=app/hello.component.spec.ts

libs/akita-ng-effects/src/lib/module-manager.service.ts
```diff
@@ -0,0 +1,46 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class ModuleManager implements OnDestroy {
+  effectInstanceSources: Type<any>[] = [];
+  destroyEffects$ = new Subject();
+
+  constructor(private actions$: Actions) {}
+
+  addEffectInstance(effectInstance) {
```
NetanelBasal, line 15: Please add types to any argument in this file

NetanelBasal:
@yackinn, you don't need a decorator. The `ngOnDestroy` hook will be called in this case.

NetanelBasal:
I had another thought. There is really nothing specific to Akita in this library. I think it'll be better to move it to `ngneat/effects` to be used with Akita, plain services, or any other approach.

WDYT?

yackinn:
> I had another thought. There is really nothing specific to Akita in this library. I think it'll be better to move it to `ngneat/effects` so that it can be used with Akita, plain services, or any other approach.
> 
> WDYT?

If you ask me effects should be part of state management. 
People that want to checkout Akita should get a sense for a complete solution wihout the need for installing some part from another project. 
It should also be documented in the Akita docs as effects belong to state management.
I also want to integrate it more with the redux dev tools.
For my taste it's too detached from the redux logs. Every action should be visible in the dev tools with their respective payload. It helps immensly for debugging.

What do you think?

libs/akita-ng-effects/src/__tests__/actions.spec.ts
yackinn: I did the same thing in webstorm and it failed.
I'll fix it

yackinn:
> @yackinn, you don't need a decorator. The `ngOnDestroy` hook will be called in this case.

I need to pass an instruction to the ngOnDestroy to unsubscrube from all subscriptions inside the class.
How would you implement unsubscribing internally when ngOnDestroy is triggered?

NetanelBasal:
> What do you think?

Yes, I agree. Let us add the integration for the devtools and schematics.

NetanelBasal:
I'll let you know about the component's providers issue. I'm busy right now, so let's leave it to the end.

NetanelBasal:
@yackinn when you finish, please merge `v6` branch into yours.

libs/akita-ng-effects/src/__tests__/actions.spec.ts
yackinn: The tests won't pass correctly using the synchronous approach.

Bildschirmfoto 2020-12-20 um 18 40 41

libs/akita-ng-effects/src/lib/module-manager.service.ts
yackinn: Types added.

yackinn:
> @yackinn when you finish, please merge `v6` branch into yours.

Branch is merged and pushed.

yackinn:
I'm thinking about what's the best solution to implement the logging of actiions with their payload in the redux dev tools.

The current state is the action decorator won't take any payload that can be logged with the action type. We can take care of this.

I don't want to overcomplicate the api.

I might have a proposal.
The action decorator can take an Action object created by the createAction function. 
The logAction function can take a StoreSnapshotAction instead of a string.

We can get the payload from the function arguments of the method that the action decorator is attached to.

We shouldn't log anything before the "reducer" in the state has performed the state update.

That means we need one action decorator for each action that should be posted in the redux dev tools.

For default actions it can just be the type update and the payload from the method arguments.


What should happen if you dispatch an action in the effects?
Would you need to trigger the store update excplicitly again with calling a action annotated method inside the store?

I guess this needs some thought.


I already tried to implement the updated action decorator and getting the payload from the class method. Logging the action with properties or payload in the redux dev tools works smoothly.

libs/akita-ng-effects/src/__tests__/actions.spec.ts
NetanelBasal: It doesn't make sense to me. I'll check it after the merge.

NetanelBasal:
I think we should provide a way to hook into the `akitaDevtools` function and dispatch custom actions. We can pass an optional `actions` stream that we can subscribe to and call `devTools.send`.

libs/akita-ng-effects/src/__tests__/actions.spec.ts
NetanelBasal: Which font are you using?

yackinn: Should the next function be invoked before subscribing and asserting?

yackinn: I created that monospaced font myself.

yackinn:
That seems feasible.
What about the action decorator? 
There are many actions in the redux dev tools that only provide a generic type such as "update" without any payload.

That's not really usefull I guess.

NetanelBasal:
You can trigger actions that don't update the state. In this case, you want to see the action and the payload.

yackinn:
That's true. But that's just one use case.
The loadBooksSuccess should trigger a state update. With the current setup that woul trigger two actions in the dev tools.
1. type: loadBooksSuccess; payload: .../ state difference: -
2. type: update: payload: -	/ state difference: ...

NetanelBasal:
You are right.

NetanelBasal:
I think we should rethink the functionality of the actions in general. I don't mind to have breaking changes as we're going to release v6.

yackinn:
I agree.
What about something like this?

// books.actions.ts
export const loadBooks = createAction("[Books] Load Books")
export const loadBooksSucceess = createAction("[Books] Load Books Success", props<{books: Books}>())

// books.effects.ts
@Effect()
loadBooks$ = this.actions.pipe(
	ofType(loadBooks),
	switchmap(_ => ...api.loadBooks().pipe(
		map(books => loadBooksSuccess({books}))
	))
)

// books.store.ts
@Reducer( loadBooksSuccess, ... )
updateBooks(books: Books) {
	this.store.update(state => {...state, books})
}

NetanelBasal:
I don't want to mix Redux with Akita. We already have actions, and I prefer not to bring the concept of reducers. 

I think we should keep it simple and transparent for the consumer. It should be handle somehow in the core Actions class.

NetanelBasal:
We can use https://github.com/datorama/akita/blob/master/libs/akita/src/lib/actions.ts#L22

yackinn:
I already did that. It works.
```ts
// public API for custom actions. Custom action always wins
export function logAction(action: StoreSnapshotAction) {
  setAction(action);
  customActionActive = true;
}
```
We need to distinguish between different action types though.
The effect shouldn't be the spot where dev tools logging is triggered from.
I think triggering the dev tool update needs to be issued by the store when the state is update by a certain action.

NetanelBasal:
https://discord.gg/GCKBpYnw

yackinn:
Logging for effects is implemented. 
Please check.

Issue #612, opened 2021-02-02
What does `dispatch?: boolean` of an effect cause exactly?


juri-sinitson:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[x] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

On https://datorama.github.io/akita/docs/angular/effects and in the file `node_modules\@datorama\akita-ng-effects\lib\types.d.ts ` is documentaton for `dispatch?: boolean` missing. 
What does `dispatch?: boolean` cause exactly?
My effects didn't work for me until I found that `dispatch` hast to be undefined instead of `false` in my case, say just skipped.

## Expected behavior

`dispatch?: boolean` is documented at least in one of the sources above and ideally also on https://datorama.github.io/akita/docs/best-practices

## Minimal reproduction of the problem with instructions
Just go to https://datorama.github.io/akita/docs/angular/effects or open `node_modules\@datorama\akita-ng-effects\lib\types.d.ts`

## What is the motivation / use case for changing the behavior?
To work with akita optimally

## Environment
Angular version: ~10.1.0
Akita version: ^6.0.0

NetanelBasal:
If you want to return a new action from the effect, you can use `{ dispatch: true }`, and it'll dispatch the returned action for you.

yackinn:
That's correct.

What's  your exact setup?
If this issue still persists can you show a stackblitz or code sample?
You should be using no flag or set ```{dispatch: true}``` if you are returning one or morge actions from your effect.

https://github.com/datorama/akita/blob/e6f0b5d83684e9ad84ddbf065c6c571912b2c50c/libs/akita-ng-effects/src/lib/module-manager.service.ts#L35-L36

juri-sinitson:
The issue is the document one, not the technical one. It works fine for me now, but it took a couple of hours for me to figure out that the beginning effect should not have `dispatch` set to false. I think, it's a mistake of a state management newbe.

That's why I expect `dispatch` to be documented on https://datorama.github.io/akita/docs/angular/effects and in `node_modules\@datorama\akita-ng-effects\lib\types.d.ts`.

Now hen I know, what `dispatch` does, I can also improve the documentation myself.
Should I just do it?

> What's your exact setup?
> If this issue still persists can you show a stackblitz or code sample?
[...]

PR #616, opened 2021-02-05
docs(code): update comments of dispatch property of EffectOptions

juri-sinitson:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #612 

## What is the new behavior?
Updating comments of dispatch property of EffectOptions

- [ ] Yes
- [x] No

### DIFF
--- a/libs/akita-ng-effects/src/lib/types.ts
+++ b/libs/akita-ng-effects/src/lib/types.ts
@@ -1,17 +1,21 @@
 import { Subject } from 'rxjs';
 
-
 export interface EffectOptions {
+  /**
+   * This parameter dictates if the effect can dispatch an action.
+   * Set `true` if you want the effect to dispatch an action.
+   * Set to `false` or omit if the effect should not dispatch any action.
+   */
   dispatch?: boolean;
 }
 
 export interface Effect extends Subject<Action> {
   isEffect: boolean;
   name: string | null;
-  dispatchAction: boolean
+  dispatchAction: boolean;
 }
 
 export interface Action {
   type: string;
-  [key: string]: any
+  [key: string]: any;
 }


juri-sinitson:
Running tests (`yarn test:all`) and building libs (``yarn build:libs`) is successful, building apps (`yarn build:apps`) fails wth:
```
ERROR: libs/akita/src/lib/store.ts:48:14 - error NG2007: Class is using Angular features but is not decorated. Please add an explicit Angular decorator.

48 export class Store<S = any> {
                ~~~~~

libs/akita/src/lib/store.ts:48:14 - error NG2007: Class is using Angular features but is not decorated. Please add an explicit Angular decorator.

48 export class Store<S = any> {
                ~~~~~
``` 
I don't think this error is relevant to my change in this pull request.

libs/akita-ng-effects/src/lib/types.ts
```diff
@@ -1,17 +1,22 @@
 import { Subject } from 'rxjs';
 
-
 export interface EffectOptions {
```
NetanelBasal, line 3: Thanks. Few comments:

- Set `true` (recommended) 
It's not recommended. It's based on your use-case. 

- or omit if you want the effect to dispatch an action.
If it's omitted or `false`, it'll not dispatch an action.


- Do not forget to set `false` for the effects which are not intended to dispatch.
`false` is the default.

juri-sinitson: This snippet makes `dispatch` to be `false` by default. I saw this snippet before I updated the comments.
https://github.com/datorama/akita/blob/e6f0b5d83684e9ad84ddbf065c6c571912b2c50c/libs/akita-ng-effects/src/lib/create-effect.util.ts#L9

I think you would like to see a documentation/comments as short as possible without redundant precautions. 
And of course without overseen mistakes.

```
This parameter dictates if the effect can dispatch an action. Set `true` if you want the effect to dispatch an action. Set to `false` or omit if the effect should not dispatch any action.
```

Is this formulation ok for you?

NetanelBasal: LGTM

juri-sinitson: The comment is now corrected

PR #622, opened 2021-02-19
docs: update docs of effects adding examples

juri-sinitson:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Simple examples of an effect chain with side effects were missing. One needs hours to get the effects chain to work.

Issue Number: N/A

## What is the new behavior?

With simple examples one can setup and test them quickly and build the productive code on them. Especially the example with HTTP-Request side effect must be very helpful.

- [ ] Yes
- [x] No

### DIFF
--- a/docs/docs/angular/effects.mdx
+++ b/docs/docs/angular/effects.mdx
@@ -31,8 +31,348 @@ import { ProductEffects } from './path/to/effects';
 })
 export class FeatureModule {}
 ```
+## The parameter `dispatch`
+This parameter dictates if the effect can dispatch an action. Set `true` if you want the effect to dispatch an action. Set to `false` or omit if the effect should not dispatch any action.
 
-## Setup Actions
+## Simple Example: a counter with logging as side effect
+Let's crete a simple counter, counting up the state and writing some info to console as a side effect.
+### Setup service
+```ts title="counter.service.ts"
+import { Injectable } from '@angular/core';
+import { Actions } from '@datorama/akita-ng-effects';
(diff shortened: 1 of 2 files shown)


yackinn:
The example isn't a probable use case for an effect. It shows how to do implement effects but it probably confuses a lot of users that are new to effects. I like to stick to a more basic example. The idea is to give a short introduction on how to use effects without needing to read too many instructions or too much code examples. It needs to be conicse.

juri-sinitson:
To leave just the effect files of my examples and put the rest to a runable stackblitz example, would it be an alternative you consider?
Would you consider creating a subpage for more examples as an alternative?

I tried to follow the aim that the official documentation has enough examples. Thus one does not need to search in 3rd sources or struggle around.

I tried to make my examples as simple as possible:
The first one is an effect chain with a counter an a `console.log` as a side effect. The next one is counter with `console.log` and an HTTP-Request as side effects. Because I put all the files needed to run, the examples may feel complicated or just blown. 

With my examples I tried to follow the principles vue.js, react, rxjs and partially angular and ngrx: provide code examples in the official documentation which one can copy, paste, run instantly an begin to transform it to productive code immediately. Without the need to read in in the topics not needed at the moment (in this case main navigation). And without the need to search for simple or alternative examples in the 3rd sources like stackoverflow threads and medium.com stories.

You can close this PR if you don't consider any proposition and any alternative I made above. 
I respect you decision.

Issue #619, opened 2021-02-14
Problem on undo with StateHistoryPlugin  when property is added on entity


lilas78:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

I'm having an entity store, and I use StateHistoryPlugin to handle undo/redo. It works fine, except for one case :

an entity gets a new property

- What I expect : when I do an undo on this entity, I expect to have the entity its in old state, so without the new property

- What I have : after the undo, the entity still has this property

To Reproduce :
[stackblitz](https://akita-books-dejkbp.stackblitz.io)
Click on edit button : this will add a comment on first book
Click on undo => the comment is not removed

See #618 

## Environment

<pre><code>
Angular version:~11.0.4

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ x] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Platform:  Windows

</code></pre>

NetanelBasal:
In this case, the following should be `replace`, not `update`:

https://github.com/datorama/akita/blob/b3341a12794ad0763f098959c04d26a378536262/libs/akita/src/lib/plugins/plugin.ts#L76

lilas78:
Is this issue goind to be fixed ?

NetanelBasal:
I'm sorry, I'm too busy right now. As I mention, it should be a quick fix. You can create a PR.

theorlovsky:
@NetanelBasal I'm digging into another problem. You can't undo removing an entity, here's the demo:
https://stackblitz.com/edit/angular-ivy-akita-undo-delete?devtoolsheight=33&file=src/app/state/service.ts

I'm planning to open a PR today. I also assume it's a related issue, so I can look into this one as well.

NetanelBasal:
Cool, thanks!

theorlovsky:
My problem with undoing an entity removing is not related to current issue, so I'll open a new one.

PR #635, opened 2021-03-03
fix: 🐛 change the update method in the StateHistoryPlugin

theorlovsky:
instead of updating state we should be replacing it

✅ Closes: #619

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

On undo/redo a store is being updated (merged), not replaced.

Issue Number: #619 

## What is the new behavior?

On undo/redo a store is being replaced.

- [ ] Yes
- [x] No

## Other information

I had to update the `EntityStore`'s `replace` method because of getting an error, something like:
> Cannot assign to 'id' because it is a read-only property

### DIFF
--- a/libs/akita/src/lib/plugins/plugin.ts
+++ b/libs/akita/src/lib/plugins/plugin.ts
@@ -1,3 +1,4 @@
+import { EntityStore } from '../entityStore';
 import { QueryEntity } from '../queryEntity';
 import { Query } from '../query';
 import { filterNil } from '../filterNil';
@@ -42,7 +43,7 @@ export abstract class AkitaPlugin<State = any> {
     }
 
     if (property) {
-      return this.getQuery().select(state => getValue(state, this.withStoreName(property)));
+      return this.getQuery().select((state) => getValue(state, this.withStoreName(property)));
     }
 
     return this.getQuery().select();
@@ -71,17 +72,23 @@ export abstract class AkitaPlugin<State = any> {
   }
 
   /** This method is responsible for updating the store or one entity; it can be the whole store or one entity. */
-  protected updateStore(newState, entityId?, property?: string) {
+  protected updateStore(newState, entityId?, property?: string, replace = false) {
     if (this.isEntityBased(entityId)) {
-      this.getStore().update(entityId, newState);
+      const store = this.getStore() as EntityStore;
+
+      replace ? store.replace(entityId, newState) : store.update(entityId, newState);
     } else {
       if (property) {
-        this.getStore()._setState(state => {
-          return setValue(state, this.withStoreName(property), newState);
+        this.getStore()._setState((state) => {
+          return setValue(state, this.withStoreName(property), newState, true);
         });
+
         return;
       }
-      this.getStore()._setState(state => ({ ...state, ...newState }));
+
+      const nextState = replace ? newState : (state) => ({ ...state, ...newState });
+
+      this.getStore()._setState(nextState);
     }
   }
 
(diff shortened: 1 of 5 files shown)


libs/akita/src/lib/plugins/plugin.ts
```diff
@@ -71,17 +72,21 @@ export abstract class AkitaPlugin<State = any> {
   }
 
   /** This method is responsible for updating the store or one entity; it can be the whole store or one entity. */
-  protected updateStore(newState, entityId?, property?: string) {
+  protected updateStore(newState, entityId?, property?: string, replace = false) {
```
NetanelBasal, line 75: Can you please change the params to one object? it'll not scale anymore

libs/akita/src/lib/setValueByString.ts
```diff
@@ -6,7 +6,7 @@ import { isObject } from './isObject';
  * @example
  * setValue(state, 'todos.ui', { filter: {} })
  */
-export function setValue(obj: any, prop: string, val: any) {
+export function setValue(obj: any, prop: string, val: any, replace = false) {
```
NetanelBasal, line 9: Same here

libs/akita/src/lib/plugins/plugin.ts
```diff
@@ -71,17 +72,21 @@ export abstract class AkitaPlugin<State = any> {
...
-        this.getStore()._setState(state => {
-          return setValue(state, this.withStoreName(property), newState);
+        this.getStore()._setState((state) => {
+          return setValue(state, this.withStoreName(property), newState, true);
         });
+
         return;
       }
-      this.getStore()._setState(state => ({ ...state, ...newState }));
+
```
NetanelBasal, line 88:
```ts
const state = replace ? newState : ((state) => ({ ...state, ...newState }))
this.getStore()._setState(start)
```

NetanelBasal (approved):

libs/akita/src/lib/plugins/plugin.ts
theorlovsky: yeah, I was thinking about it, but it will introduce a breaking change. is it ok?

NetanelBasal: This is a private method. Can you see if we use it in a different package, please?

theorlovsky: this method is being used in custom plugins

NetanelBasal: Ok, no need for now. Thanks

libs/akita/src/lib/setValueByString.ts
theorlovsky: this function is also available in the public api, so it's potentially a breaking change

libs/akita/src/lib/plugins/plugin.ts
theorlovsky: fixed

lilas78:
Since the correction in EntityStore's replace method , the object given to the replace method doesn't keep its prototype (due to replaced[id] = { ...newState, [this.idKey]: id };) 

When we get back this object from the store, we can't test the prototype anymore.

NetanelBasal:
@lilas78 are you working with classes?

lilas78:
I have a unit tests that test object equality after updating the store. This test fails with the new version of Akita.  That's why I pointed out this change.
I have to change my test to make it work.

NetanelBasal:
But the update method is also immutable.

lilas78:
Yes that's right. 
It would have been nice to be able to rely on object's prototype after querying the store. 
I will just change my test for the moment.
Thank you

PR #636, opened 2021-03-09
Build: cleanup dependencies and add yarn workspaces

matheo:
## PR Checklist

## PR Type

What kind of change does this PR introduce?}

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

The repo has some inconsistent versions of packages, and some redundant ones

## What is the new behavior?

I've cleaned up the repeated and unnecessary versions, and updated most packages to the latest available

- [ ] Yes
- [x] No

## Other information

The last commit also adds [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) to avoid issues in the package's dependencies, as the main package.json was compiling `akita` with `"schematics-utilities": "2.0.2"` but it was being published with `"^1.1.1"`


libs/akita/src/lib/store.ts
```diff
@@ -45,6 +46,7 @@ interface StoreSnapshot<S> {
  *   }
  * }
  */
+@Injectable()
```
NetanelBasal, line 49: Akita isn't coupled to Angular.

libs/akita/package.json
```diff
@@ -28,7 +28,7 @@
   "license": "Apache-2.0",
   "author": "Netanel Basal",
   "dependencies": {
-    "schematics-utilities": "^1.1.1"
+    "schematics-utilities": "^2.0.2"
```
NetanelBasal, line 31: This is a breaking change. Did you check that the schematics work as before?

package.json
```diff
@@ -7,6 +7,9 @@
     "npm": "please-use-yarn",
     "yarn": ">= 1.22.0"
   },
+  "workspaces": [
```
NetanelBasal, line 10: Why we need it here? Everything is installed in the root.

libs/akita/package.json
matheo: as mentioned in the PR, this is a bug because the library was compiled with `2.0.2` and published with `1.1.1`, fortunately the used API seems consistent because this could caused problems before.

package.json
matheo: `Nx` installs everything in the root, but doesn't validate the versions installed in the root and the one defined in the library `package.json`, so `yarn workspaces` fixes this installing the dependencies from the library instead, as we did in `spectator`, it's helpful here too.

libs/akita/src/lib/store.ts
matheo: I wonder the implications of this: https://angular.io/guide/migration-injectable

package.json
```diff
@@ -71,43 +74,40 @@
...
+    "@angular/forms": "11.2.4",
+    "@angular/platform-browser": "11.2.4",
+    "@angular/platform-browser-dynamic": "11.2.4",
+    "@angular/router": "11.2.4",
     "@netbasal/content-loader": "^1.1.0",
     "faker": "^4.1.0",
-    "helpful-decorators": "^2.0.6",
-    "lodash": "4.17.20",
-    "rxjs": "6.5.5",
-    "schematics-utilities": "2.0.2",
```
matheo, line 86: @NetanelBasal here, the `schematics-utilities` to build the libraries were different to the package version.

matheo:
I will redo this PR without the dependencies updates
I found that `@angular/compiler-cli` `ngcc` introduced the mandatory `compileUndecoratedClassesWithAngularFeatures` from v10, and it means that for some reason, `"ng-packagr": "^9.1.0"` is not using the `"@angular/compiler-cli": "11.0.5"` defined in the `package.json`, so it's better to leave that bomb quiet :man_shrugging:

matheo:
I just realized that normal `build` doesn't work, but `build --prod` does, because Ivi is disabled
Maybe the non-working npm scripts could be removed and leave only the working ones.

PR #663, opened 2021-04-26
feat(store): add set method to akita store

### LINKED ISSUE #661: Store Update - merges current and previous state values.
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[x] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

When calling store.update only the specified properties are being replaced.

If I have two possible states I want to represent in the Todos store 

```
export type TodosState = TodosBasicState | TodosAltState;

interface TodosBasicState {
  _tag: "todos-basic-state";
  uniqueToBasicObject: {
    info: string;
  };
  basicInfo: string;
}

interface TodosAltState {
  _tag: "todos-alt-state";
  altInfo: string;
}
```

I intialize my todos state to a TodosBasicState

```
const initialState: TodosState = {
  _tag: "todos-basic-state",
  uniqueToBasicObject: { info: "test" },
  basicInfo: "Basic Info"
};
```

When I update the state using the following method I get a polluted state object that is a combination of the old state and the new values.

```
this.todosStore.update((state) => ({
      _tag: "todos-alt-state",
      altInfo: "This is some alt info.",
    }));
```

State object as json before update:

```
{
  "_tag": "todos-basic-state",
  "uniqueToBasicObject": {
    "info": "test"
  },
  "basicInfo": "Basic Info"
}
```

State object as json after update:
```
{
  "_tag": "todos-alt-state",
  "uniqueToBasicObject": {
    "info": "test"
  },
  "basicInfo": "Basic Info",
  "altInfo": "This is some alt info."
}
```

As you can see the the _tag and altInfo properties have been updated as expected but the rest of the state has not been removed from the previously.

The following line is the cause in store.ts. The current and new state are both being combined in a spread operation. I'm sure this is being done for a reason I'm ignorant of but is this side effect intentional?

https://github.com/datorama/akita/blob/721243c814b8835a0bec113a686c5a1400dc2f2a/libs/akita/src/lib/store.ts#L256

## Expected behavior

I would expect that when calling the update method I get only the specified properties in the state and all old properites are dropped.

Expect this:

```
this.todosStore.update((state) => ({
      _tag: "todos-alt-state",
      altInfo: "This is some alt info.",
    }));
```

to turn state from:

```
{
  "_tag": "todos-basic-state",
  "uniqueToBasicObject": {
    "info": "test"
  },
  "basicInfo": "Basic Info"
}
```

to this:
```
{
  "_tag": "todos-alt-state",
  "altInfo": "This is some alt info."
}
```

If I wanted to keep values from the previous state I would expect to have to include them in the update callback using a spread or explicitly.

e.g.
```
this.todosStore.update(state => ({
      ...state,
      _tag: "todos-alt-state",
      altInfo: "This is some alt info."
    }));
```

https://stackblitz.com/edit/akita-todos-app-pyf2xv?file=src/app/todos/state/todos.query.ts

Click update to perform the above update method and click reset to reset the store back to the initial method.

I want to be able to update the state without previous state values polluting the new state. If there is a better way to achieve this any advise is welcome?

## Environment

<pre><code>
Angular version: X.Y.Z

Browser:
- [ x] Chrome (desktop) version XX
- [ x] Chrome (Android) version XX
- [ x] Chrome (iOS) version XX
- [ x] Firefox version XX
- [ x] Safari (desktop) version XX
- [ x] Safari (iOS) version XX
- [ x] IE version XX
- [ x] Edge version XX
</code></pre>

field123:
Created a public method on the akita store that will completely replace the current state with the new provided state and not just update the value.

Closes #661

## PR Checklist

- [ x ] The commit message follows our guidelines: https://github.com/datorama/akita/blob/master/CONTRIBUTING.md#commit
- [ x ] Tests for the changes have been added (for bug fixes / features)
- [ x ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #661

## What is the new behavior?

When the `set()` method is used the current state of the store will be replaced not just { ...currentState, ...newState } merged together.

See issue #661 for more details.

- [ ] Yes
- [x] No

### DIFF
--- a/docs/docs/store.mdx
+++ b/docs/docs/store.mdx
@@ -37,13 +37,13 @@ When you want to update the store, you can call the `update()` method passing th
 
 ```ts title="session.service.ts"
 import { SessionStore } from './session.store';
- 
+
 export class SessionService {
   constructor(private sessionStore: SessionStore) {}
 
   updateUserName(newName: string) {
     this.sessionStore.update({ name: newName });
-  }  
+  }
 }
 ```
 
@@ -51,7 +51,7 @@ The second `update()` option gives you more control. It receives a `callback` fu
 
 ```ts title="session.service.ts"
 import { SessionStore } from './session.store';
- 
+
 export class SessionService {
   constructor(private sessionStore: SessionStore) {}
 
@@ -59,18 +59,104 @@ export class SessionService {
     this.sessionStore.update(state => ({
       name: newName
     }));
-  }  
+  }
 }
 ```
+
+### `set()`
+
+When you don't just want to update the current store state but replace the state completely, you can call the `set()` method passing the new `state`:
+
+```ts title="session.service.ts"
+import { SessionStore } from './session.store';
+
+export class SessionService {
+  constructor(private sessionStore: SessionStore) {}
+
+  updateUserName(newName: string) {
+    this.sessionStore.set({ name: newName });
+  }
+}
+```
+
+As with `update()` the `set()` method also has a second option which gives you more control. It receives a `callback` function, which gets the current state, and returns a new **immutable** state, which will be the new value of the store. For example:
+
+```ts title="session.service.ts"
+import { SessionStore } from './session.store';
+
+export class SessionService {
+  constructor(private sessionStore: SessionStore) {}
+
+  updateUserName(newName: string) {
+    this.sessionStore.set(state => ({
+      name: newName
+    }));
+  }
+}
+```
+
+You should use `set()` over `update()` when you want to completely replace the current state at the top level.
+
+```ts title="session.store.ts"
+type SessionState = SessionStateAuthenticated | SessionStateUnauthenticated;
+
+interface SessionStateAuthenticated {
+  _tag: 'session-state-authenticated';
+  user: User;
+}
+
+interface SessionStateUnauthenticated {
+  _tag: 'session-state-unauthenticated'
+}
+
+interface User {
+  uid: string;
+  displayName: string;
+}
+
+```
+
+```json "current state"
+{
+  "_tag": "session-state-authenticated",
+  "user": {
+    "uid": "123",
+    "displayName": "John Doe"
+  }
+}
+```
+
+Using the above state model and value of current state, if we call `this.store.update({ _tag: 'session-state-unauthenticated' })` our state will be updated to the following:
+
+```json "using the update() method"
+{
+  "_tag": "session-state-unauthenticated",
+  "user": {
+    "uid": "123",
+    "displayName": "John Doe"
+  }
+}
+```
+
+Our value for tag has been updated but the user value from our previous state has stayed, this could lead to unintended side effects so in this case `this.store.set({ _tag: 'session-state-unauthenticated' })` may be a better option. When we use the `set()` method the following is the result:
+
+```json "using the update() method"
+{
+  "_tag": "session-state-unauthenticated",
+}
+```
+
+We have replaced the previous state completely, leaving behind no unexpected state values.
+
 
 ### `setLoading()`
 
 Set the `loading` state:
 ```ts title="session.service.ts"
 import { SessionStore } from './session.store';
- 
+
 export class SessionService {
-  constructor(private sessionStore: SessionStore, 
+  constructor(private sessionStore: SessionStore,
               private http: HttpClient) {}
 
   async updateUserName(newName: string) {
@@ -78,7 +164,7 @@ export class SessionService {
     await this.http(...).toPromise();
     this.sessionStore.update({ name: newName});
     this.sessionStore.setLoading(false);
-  }  
+  }
 }
 ```
 
@@ -87,9 +173,9 @@ Set the `error` state:
 
 ```ts title="session.service.ts"
 import { SessionStore } from './session.store';
- 
+
 export class SessionService {
-  constructor(private sessionStore: SessionStore, 
+  constructor(private sessionStore: SessionStore,
               private http: HttpClient) {}
 
   async updateUserName(newName: string) {
@@ -98,7 +184,7 @@ export class SessionService {
     } catch(error) {
       this.sessionStore.setError(error);
     }
-  }  
+  }
 }
 ```
 
--- a/libs/akita/src/lib/store.ts
+++ b/libs/akita/src/lib/store.ts
@@ -226,7 +226,8 @@ export class Store<S = any> {
 
   /**
    *
-   * Update the store's value
+   * Update the store's value, only replacing the specified properties
+   *
    *
    * @example
    *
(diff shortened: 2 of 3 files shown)


docs/docs/store.mdx
```diff
@@ -37,48 +37,134 @@ When you want to update the store, you can call the `update()` method passing th
...
+  }
+}
+```
+
+### `set()`
+
+When you don't just want to update the current store state but replace the state completely, you can call the `set()` method passing the new `state`:
+
+```ts title="session.service.ts"
+import { SessionStore } from './session.store';
```
NetanelBasal, line 71: We can' use the name `set` because it's an EntityStore method.

field123: A few alternatives ideas:

`put`

The positives would be it's familiar terminology to people, it performs as expected. With the caveat that in the HTTP spec it's idempotent and this method would not be technically idempotent if a akitaPre hook performs a side effect.

`overwrite`

At a glance it feels descriptive of the functionality it performs. When paired with `update()` give an immediate intuition about the differences between them. Does not feel as familiar in the world of CRUD updating.

`setState`

Very explicit about what it does. Is close to the @internal `_setState()` method which could cause confusion. Inclusion of the word 'State' doesn't gel as well with the update(), sounds less like an alternative and more like different functionality.

My personal preference order: `overwrite` -> `put` -> `setState`.

NetanelBasal: let's go with overwrite

field123: Done.

docs/docs/store.mdx
```diff
@@ -37,48 +37,134 @@ When you want to update the store, you can call the `update()` method passing th
...
 }
 ```
 
+You should use `overwrite()` over `update()` when you want to completely replace the current state at the top level.
+
+```ts title="session.store.ts"
+type SessionState = SessionStateAuthenticated | SessionStateUnauthenticated;
+
+interface SessionStateAuthenticated {
+  _tag: 'session-state-authenticated';
```
NetanelBasal, line 104: I prefer using a simple example state and not copy-paste from your application.

docs/docs/store.mdx
```diff
@@ -37,48 +37,134 @@ When you want to update the store, you can call the `update()` method passing th
...
+{
+  "_tag": "session-state-authenticated",
+  "user": {
+    "uid": "123",
+    "displayName": "John Doe"
+  }
+}
+```
+
+Using the above state model and value of current state, if we call `this.store.update({ _tag: 'session-state-unauthenticated' })` our state will be updated to the following:
```
NetanelBasal, line 129: I don't think we need all of this. The method name is straightforward.

libs/akita/src/lib/store.ts
```diff
@@ -244,18 +245,49 @@ export class Store<S = any> {
...
+   */
+  overwrite(state: S);
+  overwrite(stateOrCallback: S | UpdateStateCallback<S>): void {
+    isDev() && setAction('Overwrite');
+    const withHookFn = (curr: S, newS: S) => this.akitaPreOverwrite(curr,  newS as S);
+    this._setState(this.prepareNewState(stateOrCallback, this._value(), withHookFn));
+  }
+
+  private prepareNewState<S>(stateOrCallback: Partial<S> | UpdateStateCallback<S>, currentState: S, withHookFn: (c: S, n: S) => S): S {
+    const constructNewState = (x: Partial<S> | UpdateStateCallback<S>, cs: S) => {
```
NetanelBasal, line 277: - Why do we need a function? 
- Why do we need to create a new function each time?

libs/akita/src/lib/store.ts
```diff
@@ -244,18 +245,49 @@ export class Store<S = any> {
...
+      if (isFunction(x)) {
+        return isFunction(this._producerFn) ? this._producerFn(currentState, x) : x(cs);
+      } else {
+        return x;
+      }
+    }
+    const resolveFinalState = (currentState: S, withHook: S): S => {
```
NetanelBasal, line 284: Same here

libs/akita/src/lib/store.ts
field123: 1. Why do we need a function?

I've used a hook function so the prepareNewState method is not bound to a specific class method e.g akitaPreOverwrite or akitaPreUpdate instead a wrapping function of the relevant class method is used in update, overwrite or any future methods. This allows for the same class class overriding of the methods yet still being able to pass them as HOF.
```const hookFn = (curr: S, newS: S) => this.akitaPreOverwrite(curr,  newS as S)``` 

Using the hook means the ordering of operations can be maintaned.

2. Why do we need to create a new function each time?

In an effort to make small single purpose functions I like to lexically scope util functions. It does have the possible performance hit but I think the clarity of being able to scroll to the bottom of a method/fn and know what's happening is worth it.

```
const newState = extractNewState(stateOrCallback, currentState);
const withHook = hookFn(currentState, newState);
return resolveFinalState(currentState, withHook);
```

This makes it clear I'm extracting the new state from my callback and current state -> apply my hook to the currentState and newState -> finally resolving the final state. I could makes these more descriptive but don't understand what the `_producerFn` is or why your checking if the state is an object.

You can also pluck out these small functions and raising their scope if you have a need for them elsewhere.

The alternative would be

```
let newState;
if (isFunction(stateOrCallback)) {
 newState = isFunction(this._producerFn) ? this._producerFn(currentState, stateOrCallback) : stateOrCallback(currentState);
} else {
 newState = stateOrCallback;
};
const withHook = hookFn(currentState, newState);
return isPlainObject(currentState) ? withHook : new (currentState as any).constructor(withHook);
```

While this can be made more clear through comments it could be tricky to understand what the fn/method is doing.

If the nest functions are a concern from performance we could put the functions in private methods, extract them out of the class or of course just remove them and have all the code blocked together.

```
// Option 1
// Inside the Store class

private extractNewState(x: Partial<S> | UpdateStateCallback<S>, cs: S) {
    if (isFunction(x)) {
      return isFunction(this._producerFn) ? this._producerFn(cs, x) : x(cs);
    } else {
      return x;
    }
  }

  private resolveFinalState(currentState: S, withHook: S): S {
    return isPlainObject(currentState) ? withHook : new (currentState as any).constructor(withHook);
  }

// Option 2

// Before or after the Store class def

const extractNewState = <S>(x: Partial<S> | UpdateStateCallback<S>, cs: S, _producerFn: (state: any, fn: any) => any) => {
  if (isFunction(x)) {
    return isFunction(_producerFn) ? _producerFn(cs, x) : x(cs);
  } else {
    return x;
  }
};

const resolveFinalState = <S>(currentState: S, withHook: S): S => {
  return isPlainObject(currentState) ? withHook : new (currentState as any).constructor(withHook);
};
```

That's my reasoning, happy to do what ever fits best with the project. 👍

File: docs/docs/store.mdx (hunk 2 above)
Comment (field123) [reply docs/docs/store.mdx:104]: This was a much reduced version of the original, I though that session state was a good simple example and "real world" enough to be useful. My preference when reading documentation has always been simplified "real world" over slightly contrived.

I could simplify the example further to

```
type SessionState = AuthenticatedState | UnauthenticatedState;

interface AuthenticatedState {
  _tag: 'authenticated';
  uid: string;
}

interface UnauthenticatedState {
  _tag: 'unauthenticated'
}
```

or even do something like

```
type SessionState = TokenState | InitialState;

interface TokenState {
  name: string;
  token: string;
}

interface InitialState{
  name: string;
}
```

File: docs/docs/store.mdx (hunk 3 above)
Comment (field123) [reply docs/docs/store.mdx:129]: I was concerned that newer users may be unsure when to use each method, but it could be way to verbose.

File: docs/docs/store.mdx (hunk 2 above)
Comment (NetanelBasal) [reply docs/docs/store.mdx:104]: I prefer the latter

File: docs/docs/store.mdx (hunk 3 above)
Comment (NetanelBasal) [reply docs/docs/store.mdx:129]: The `overwrite` name is clear enough.

File: libs/akita/src/lib/store.ts (hunk 4 above)
Comment (NetanelBasal) [reply libs/akita/src/lib/store.ts:277]: My question was why we need that `constructNewState` and `resolveFinalState` will be local functions?

File: docs/docs/store.mdx (hunk 2 above)
Comment (field123) [reply docs/docs/store.mdx:104]: I'll change to the second option.

File: docs/docs/store.mdx (hunk 3 above)
Comment (field123) [reply docs/docs/store.mdx:129]: I'll make the change.

File: libs/akita/src/lib/store.ts (hunk 4 above)
Comment (field123) [reply libs/akita/src/lib/store.ts:277]: I find this: 

```
const extractNewState = (x: Partial<S> | UpdateStateCallback<S>, cs: S) => {
      if (isFunction(x)) {
        return isFunction(this._producerFn) ? this._producerFn(cs, x) : x(cs);
      } else {
        return x;
      }
};

const inCaseClassResolveState= (currentState: S, withHook: S): S => {
   return isPlainObject(currentState) ? withHook : new (currentState as any).constructor(withHook);
};

const newState = extractNewState(stateOrCallback, currentState);
const withHook = hookFn(currentState, newState);
return inCaseClassResolveState(currentState, withHook);
```

To be more immediately understandable than:

```
let newState;
if (isFunction(stateOrCallback)) {
 newState = isFunction(this._producerFn) ? this._producerFn(currentState, stateOrCallback) : stateOrCallback(currentState);
} else {
 newState = stateOrCallback;
};
const withHook = hookFn(currentState, newState);
return isPlainObject(currentState) ? withHook : new (currentState as any).constructor(withHook);
```

If future me or another developer comes along and sees the line of code `isPlainObject(currentState) ? withHook : new (currentState as any).constructor(withHook);` it may not make sense without diving in and understanding it. But if the block of code is wrapped in a function called `inCaseClassResolveState` you get some understanding straight away of what the operation is doing.

With the added benefit of not mutating (with variable reassignment) the newState variable once declared. e.g. using `const` and not `let`.

I'm happy to just change the class method to match below if you prefer?

```
private prepareNewStateAlt<S>(stateOrCallback: Partial<S> | UpdateStateCallback<S>, currentState: S, hookFn: (curr: Readonly<S>, newS: Readonly<S>) => S): S {
    let newState;
    if (isFunction(stateOrCallback)) {
      newState = isFunction(this._producerFn) ? this._producerFn(currentState, stateOrCallback) :    stateOrCallback(currentState);
    } else {
      newState = stateOrCallback;
    }
    const withHook = hookFn(currentState, newState);
    return isPlainObject(currentState) ? withHook : new (currentState as any).constructor(withHook);
}
```

NetanelBasal: Add comments please, we don't need to create redundant functions.

field123: Changes done.

tony:
@field123 Can you rebase?

PR #670, opened 2021-05-02
fix(akita-ng-entity-service): loader late subscriber

### LINKED ISSUE #668: NgEntityService Loader Misses First Emission When using Async Pipe & NgOnInit
## I'm submitting a...
<pre><code>
[x] Bug report 
[ ] Performance issue
[ ] Feature request
[ ] Support request
[ ] Other... Please describe:
</code></pre>

## Current behavior

Based on documentation*, I should be able to call an NgEntityService .get() in my components ngOnInit() function and subscribe to an NgEntityServiceLoader observable in the component template using async pipe.

[1] Docs: https://datorama.github.io/akita/docs/angular/entity-service#entity-service-loader

However, since the dispatcher in NgEntityServiceLoader is a subject, and the call to NgEntityService .get() occurs before the NgEntityServiceLoader observable is subscribed to via async pipe, the first emission is always missed.

## Expected behavior

Subscribing to an NgEntityServiceLoader observable should return the current state of the observable / loader.

I created an example close to the documentation. Caveat, I added an interceptor to delay the HTTP request to clearly show this behavior.

[2] Reproduction: https://stackblitz.com/edit/akita-loader-async-pipe-problem?file=src/app/hello.component.ts

The use case is the same as the documentation. Subscribe to .get() in your components ngOnInit, and subscribe to one of the loader's observables in the component's template via async pipe. 

Today from a client perspective (no changes to lib), you could remedy this problem by creating your async pipe outside your component and passing the loaders value in as an input, passing the value of the loader to a Replay/BehaviorSubject, subscribing directly in NgOnInit, etc...  None of these methods are as convenient / idiomatic as the example in the documentation.

## Environment

<pre><code>
@datorama/akita: 6.1.3
@datorama/akita-ng-entity-service: 6.0.0
Angular version: 11.2.12
</pre></code>

Browser:
- [x] Chrome (desktop) version 90.0.4430.93 (Official Build) (64-bit)
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:

Others:

Can I make a PR for this? -- I believe it could be remedied by updating NgEntityServiceLoader's dispatcher to a ReplaySubject. Although I'm not sure if there's a case where that would cause negative regression. Could this possibly be toggled (ReplaySubject or Subject) behind a configuration property?

Everduin94:
Added Configuration to enable NgEntityServiceLoader to utilize a ReplaySubject

fixes #668

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #668

## What is the new behavior?

An option to use a `ReplaySubject` over a `Subject` for the `NgEntityServiceLoader` dispatcher. This allows late subscribers (e.g. components that call `get()` via `ngOnInit` and subscribe to `get$` via async pipe in the template) to receive the latest emission on subscription.

- [ ] Yes
- [x] No

### DIFF
--- a/libs/akita-ng-entity-service/src/lib/ng-entity-service.config.ts
+++ b/libs/akita-ng-entity-service/src/lib/ng-entity-service.config.ts
@@ -1,6 +1,7 @@
 import { InjectionToken } from '@angular/core';
 import { NgEntityServiceParams } from './types';
 import { HttpMethod } from './ng-entity-service-notifier';
+import { Event } from './ng-entity-service.loader';
 import { isObject } from '@datorama/akita';
 import { Subject } from 'rxjs';
 
@@ -13,7 +14,7 @@ export interface NgEntityServiceGlobalConfig {
     PUT: HttpMethod;
     DELETE: HttpMethod;
   }>;
-  connector?: () => Subject<{ method: HttpMethod; loading: boolean; storeName: string; entityId?: any }>;
+  connector?: () => Subject<Event>;
 }
 
 export const NG_ENTITY_SERVICE_CONFIG = new InjectionToken<NgEntityServiceGlobalConfig>('NgEntityServiceGlobalConfig');
--- a/libs/akita-ng-entity-service/src/lib/ng-entity-service.config.ts
+++ b/libs/akita-ng-entity-service/src/lib/ng-entity-service.config.ts
@@ -12,6 +12,7 @@ export interface NgEntityServiceGlobalConfig {
     PUT: HttpMethod;
     DELETE: HttpMethod;
   }>;
+  replayLoader?: boolean;
 }
 
 export const NG_ENTITY_SERVICE_CONFIG = new InjectionToken<NgEntityServiceGlobalConfig>('NgEntityServiceGlobalConfig');
@@ -22,8 +23,8 @@ export const defaultConfig: NgEntityServiceGlobalConfig = {
     POST: HttpMethod.POST,
     PATCH: HttpMethod.PATCH,
     PUT: HttpMethod.PUT,
-    DELETE: HttpMethod.DELETE
-  }
+    DELETE: HttpMethod.DELETE,
+  },
 };
 
 export function mergeDeep(target, ...sources) {
@@ -45,7 +46,7 @@ export function mergeDeep(target, ...sources) {
 }
 
 export function NgEntityServiceConfig(config: NgEntityServiceParams = {}) {
-  return function(constructor) {
+  return function (constructor) {
     if (config.baseUrl) {
       constructor['baseUrl'] = config.baseUrl;
     }
--- a/libs/akita-ng-entity-service/src/lib/ng-entity-service.config.ts
+++ b/libs/akita-ng-entity-service/src/lib/ng-entity-service.config.ts
@@ -2,6 +2,7 @@ import { InjectionToken } from '@angular/core';
 import { NgEntityServiceParams } from './types';
 import { HttpMethod } from './ng-entity-service-notifier';
 import { isObject } from '@datorama/akita';
+import { Subject } from 'rxjs';
 
 export interface NgEntityServiceGlobalConfig {
   baseUrl?: string;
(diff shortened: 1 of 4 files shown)


libs/akita-ng-entity-service/src/lib/ng-entity-service.spec.ts
```diff
@@ -277,7 +277,7 @@ describe('NgEntityService', () => {
           TestServiceWithInlineConfig,
           {
             provide: NG_ENTITY_SERVICE_CONFIG,
-            useValue: {} as NgEntityServiceGlobalConfig,
+            useValue: { replayLoader: true } as NgEntityServiceGlobalConfig,
```
Everduin94, line 280: I set the replayLoader to true for all 'get' tests. I assume this isn't what we want but I only placed replayLoader on the global config. 
- Should I add replayLoader to service level configs as well?
- Or just modify the global config for my single test?

libs/akita-ng-entity-service/src/lib/ng-entity-service.loader.ts
```diff
@@ -1,14 +1,15 @@
...
 import { isFunction } from '@datorama/akita';
 import { HttpMethod } from './ng-entity-service-notifier';
+import { NG_ENTITY_SERVICE_CONFIG } from '@datorama/akita-ng-entity-service';
 
 type Event = { method: HttpMethod; loading: boolean; storeName: string; entityId?: any };
 
 @Injectable({ providedIn: 'root' })
 export class NgEntityServiceLoader {
-  private dispatcher = new Subject<Event>();
+  private dispatcher = inject(NG_ENTITY_SERVICE_CONFIG)?.replayLoader ? new ReplaySubject<Event>() : new Subject<Event>();
```
Everduin94, line 12: Is a function preferred here, or is this okay?

Everduin94:
I wasn't sure what to add to the documentation since the property changes an underlying class implementation. Should I explain the property or just mention it in the config properties here? https://datorama.github.io/akita/docs/angular/entity-service#options

libs/akita-ng-entity-service/src/lib/ng-entity-service.loader.ts
NetanelBasal: I prefer a factory:

`{ connector: () => new ReplaySubjec(1) }`

Everduin94: Sorry I'm familiar with factories but not 100% following in this context.

Something like this?

```typescript
const createDispatcherImpl = (config: NgEntityServiceGlobalConfig) => ({
  connector: () => config.replayLoader ? new ReplaySubject(1) : new Subject()
});

@Injectable({ providedIn: 'root' })
export class NgEntityServiceLoader {

  private dispatcher = createDispatcherImpl(inject(NG_ENTITY_SERVICE_CONFIG)).connector();
  // ...
}
```

NetanelBasal: No. I mean that instead of the boolean `replayLoader`.

```ts
{ connector: () => new ReplaySubjec(1) }
```

Everduin94: 😅 Ah that makes sense, my bad! I replaced replayLoader boolean with connector subject factory and added that property to the documentation.

libs/akita-ng-entity-service/src/lib/ng-entity-service.config.ts
```diff
@@ -12,6 +13,7 @@ export interface NgEntityServiceGlobalConfig {
     PUT: HttpMethod;
     DELETE: HttpMethod;
   }>;
+  connector?: () => Subject<{ method: HttpMethod; loading: boolean; storeName: string; entityId?: any }>;
```
NetanelBasal, line 16: Why you are not using the Event type?

Everduin94: Fixed. -- Event wasn't being exported and shares the same name as Event from the DOM library. So I wasn't sure if there was a reason I wasn't aware of that it should only be used within the NgEntityServiceLoader. I actually put Event at first but then second-guessed myself 😅.

PR #745, opened 2021-09-09
feat(akita): more typesafe EntityState

RobinVdBroeck:
This deletes the {[key: string]: any} from the definiton of EnityState.
This is breaking change, because users of akita might have failing
typechecking now. (However, this should have failed either way, since it
was not correct.)

See #744 for the reasoning behind this change.

## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/datorama/akita/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: Typesafety

Issue Number: #744

- [X] Yes
- [ ] No

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -563,6 +563,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "RobinVdBroeck",
+      "name": "Robin Van den Broeck",
+      "avatar_url": "https://avatars.githubusercontent.com/u/3083785?v=4",
+      "profile": "https://robinvdb.me",
+      "contributions": [
+        "code"
+      ]
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


NetanelBasal:
Check if this code works:

```ts
export interface TodosState extends EntityState<Todo, number> {
  filter: string;
}

@StoreConfig({ name: 'todos' })
export class TodosStore extends EntityStore<TodosState> {
  constructor() {
    super({ filter: 'ALL' });
  }
}
```

RobinVdBroeck:
> Check if this code works:
> 
> ```ts
> export interface TodosState extends EntityState<Todo, number> {
>   filter: string;
> }
> 
> @StoreConfig({ name: 'todos' })
> export class TodosStore extends EntityStore<TodosState> {
>   constructor() {
>     super({ filter: 'ALL' });
>   }
> }
> ```

It does. I've also added an extra test case for this.

NetanelBasal:
When you run `npm run start` it compiles?

NetanelBasal:
There will be a problem with the `active` property. Let me know if you can handle it.

RobinVdBroeck:
> There will be a problem with the `active` property. Let me know if you can handle it.

I'm currently looking into it. I hope to have something ready by the start of next week.

RobinVdBroeck:
It looks like EntityStore expects the given EntityState to also be either a ActiveState or MultiActiveState for functions like getActive(), setActions() and updateActive(). What's the design decision behind this? Should EntityStore work without active?

NetanelBasal:
Yes, it's optional. The consumer should pass it. We can solve it by something like:

```ts
S extends { active: any } ? S['active'] : never
```

RobinVdBroeck:
> Yes, it's optional. The consumer should pass it. We can solve it by something like:
> 
> ```ts
> S extends { active: any } ? S['active'] : never
> ```
I was thinking about using never to. However, this is valid typescript:

```typescript
declare let a: never;
let b: any = a;
```
[playground](https://www.typescriptlang.org/play?ts=4.3.5&ssl=3&ssc=16&pln=1&pc=1#code/CYUwxgNghgTiAEEQBd5QFzwHYgG4hgG4AoYpVAI0yiwE94BeNQoA)

I think we can live with that. If the consumer of the api uses any, they are throwing type checking away either way.

RobinVdBroeck:
I don't get it compiling, can you take a look at it?

PR #768, opened 2021-11-07
fix: Explicit check for boolean value

field123:
akita-ng-effects: Had a situation where AngularFireAuth was a object of a custom Proxy type that when property.isEffect was called, it returned an object which evaluated as truthy passing the if statement before blowing up.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When isEffect was called on a property that does not return undefined and instead returns something non boolean but truthy the if statement passes incorrectly. 

```
property.isEffect // in my case the call resolves to Proxy { length: 0, name: '' } which is truthy and passes the check
```

```
if (property.isEffect) {
  // blows up in here because it's not an effect.
});

```

Issue Number: N/A

## What is the new behavior?

changing to an explicit check for boolean true fixes the problem

```
if (property.isEffect === true) {
  // truthy object no longer reaches here.
});
```

- [ ] Yes
- [x] No

### DIFF
--- a/libs/akita-ng-effects/src/__tests__/module-manager.service.spec.ts
+++ b/libs/akita-ng-effects/src/__tests__/module-manager.service.spec.ts
@@ -9,6 +9,7 @@ describe('Module Manager Service', () => {
   let mixedInstance;
   let mockEffectOne;
   let mockEffectTwo;
+  let truthyNoEffect;
   let noEffect;
   let actions$;
   let observable$;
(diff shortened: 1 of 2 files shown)


field123:
Hi, @NetanelBasal do you have a timeframe on publishing this to npm? Thanks.

NetanelBasal:
Today

NetanelBasal:
done

Issue #870, opened 2022-09-07
Support strict mode


corsairdnb:
### Is this a regression?

No

### Description

When trying to run Angular build or tests got following errors:

```
Error: node_modules/@datorama/akita/src/lib/entityStore.d.ts:262:83 - error TS2344: Type 'UIState' does not satisfy the constraint 'EntityState<any, any>'.

262 export declare class EntityUIStore<UIState, DEPRECATED = any> extends EntityStore<UIState> {
                                                                                      ~~~~~~~

  node_modules/@datorama/akita/src/lib/entityStore.d.ts:262:36
    262 export declare class EntityUIStore<UIState, DEPRECATED = any> extends EntityStore<UIState> {
                                           ~~~~~~~
    This type parameter might need an `extends EntityState<any, any>` constraint.


Error: node_modules/@datorama/akita/src/lib/plugins/plugin.d.ts:3:65 - error TS2344: Type 'State' does not satisfy the constraint 'EntityState<any, any>'.

............
```

Steps to reproduce:

1. npx @angular/cli new project (>=14.2.0 version expected to install)
2. npm i typescript@4.8.2 (check that typescript >=4.8 installed)
3. ng add @datorama/akita
4. ng g m session
5. ng g c session/signup
6. ng g af session/session --plain
7. add code to AppComponent: `constructor(private sessionService: SessionService) {}`
8. run `npm run test` or `npm run build`

### Please provide the exception or error you saw

```
Error: node_modules/@datorama/akita/src/lib/entityStore.d.ts:262:83 - error TS2344: Type 'UIState' does not satisfy the constraint 'EntityState<any, any>'.

262 export declare class EntityUIStore<UIState, DEPRECATED = any> extends EntityStore<UIState> {
                                                                                      ~~~~~~~

  node_modules/@datorama/akita/src/lib/entityStore.d.ts:262:36
    262 export declare class EntityUIStore<UIState, DEPRECATED = any> extends EntityStore<UIState> {
                                           ~~~~~~~
    This type parameter might need an `extends EntityState<any, any>` constraint.


Error: node_modules/@datorama/akita/src/lib/plugins/plugin.d.ts:3:65 - error TS2344: Type 'State' does not satisfy the constraint 'EntityState<any, any>'.

3 export declare type Queries<State> = Query<State> | QueryEntity<State>;
                                                                  ~~~~~

  node_modules/@datorama/akita/src/lib/plugins/plugin.d.ts:3:29
[...]

PR #1050, opened 2023-01-31
fix(akita): support TS strict mode

Den-dp:
This PR makes the public API of Akita to be consumable in a strict typescript project (and without skipLibCheck).

Refs: salesforce/akita#870

----

The first thing I've tried is to enable `strict: true` for this monorepo and fix them - but the amount of errors in such case is pretty significant 🥵

After that, I tried to fix only TS errors that I found while consuming akita@8.0.0 in my project. And such approach helped me to decrease the API surface to a couple of places, which was somehow fixed in this draft/spike #1050. 

Interestingly, #1050 already allowed me to consume akita in Angular 15 project with a strict mode (and without skipLibCheck).

But unfortunately:
~~1. I'm not confident in these TS-related changes (don't have a big project to test on)~~
**UPD: @qnku @nunoarruda confirmed that it works for them** 
~~2. I want to find a proper tsconfig configuration that will help with the reproduction of the initial issue for future contributors~~
**UPD: I'm thinking of converting `ng-playground` into strict mode and switching it to consume a lib from `dist` (the last one does not oblige us to fix all akita strict issues but only d.ts related) as a separate PR**

But any recommendations on these 2 subjects are welcome!

Also, I'm totally ok if someone wants to use my spike to fix this issue ❤️

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "./node_modules/nx/schemas/nx-schema.json",
   "tasksRunnerOptions": {
     "default": {
       "runner": "nx/tasks-runners/default",
@@ -7,7 +8,6 @@
       }
     }
   },
-  "extends": "nx/presets/npm.json",
   "npmScope": "@datorama",
   "affected": {
     "defaultBase": "master"
@@ -33,6 +33,9 @@
     }
   },
   "targetDefaults": {
+    "build": {
+      "dependsOn": ["^build"]
+    },
     "test": {
       "inputs": ["default", "^default", "{workspaceRoot}/jest.preset.js"]
     },
--- a/packages/akita/src/lib/entityStore.ts
+++ b/packages/akita/src/lib/entityStore.ts
@@ -663,7 +663,7 @@ export class EntityStore<S extends EntityState = any, EntityType = getEntityType
 }
 
 // @internal
-export class EntityUIStore<UIState, DEPRECATED = any> extends EntityStore<UIState> {
+export class EntityUIStore<UIState extends EntityState, DEPRECATED = any> extends EntityStore<UIState> {
   _akitaCreateEntityFn: EntityUICreateFn;
 
   constructor(initialState = {}, storeConfig: Partial<StoreConfigOptions> = {}) {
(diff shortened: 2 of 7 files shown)


qnku:
@Den-dp do you have any idea why the PR build fails? 

When I run `npm run build:all` with your changes I do not get any errors ...

Den-dp:
@qnku thanks for your interest!

**TL;DR**: CI fails cause of nx-related misconfiguration, so not really relevant (also I already have a fix ✔️ for that)  

Where I need some help is in testing these changes on a real code with a strict mode enabled 😉 

It's relatively simple:
```bash
# build akita and pack it
npx nx build akita
cd dist/packages/akita
npm pack

# install it in your project using full path to the datorama-akita-8.0.0.tgz
npm i C:\github\akita\dist\packages\akita\datorama-akita-8.0.0.tgz
```

-----------

But if you are really interested in "why it works on your machine and doesn't on CI" then here are my findings:
 * `npm run build:all` fails on CI because nx is not configured to build `ng-entity-service` and `ng-router-store` after `akita` (it tries to build them all in parallel which is not possible)
 * It might work on your PC cause you probably already have `akita` artifacts in `dist` folder so nx is able to build the remaining two libs (can repro it by running `npm run build:all` twice)

qnku:
@Den-dp Thanks for the quick update. 

We have a medium sized application and use the stores almost everywhere.

I can confirm that the build errors we previously got with the 8.0.0 release and Angular 15 do not occur anymore with your fix. Also found no runtime errors during my tests (which I didn't expect anyway, since you just changed the typings).

For the test I used your installation instruction, turned `skipLibChecks` off again. Strict mode has always been on.

Very happy, hoping for a release soon :-)

nunoarruda:
Hey @Den-dp, thanks for the effort 👍 

I have tested your changes in an Angular 15 project with strict mode enabled. No more type/compilation errors. My stores of type `Store` and of type `EntityStore` are working fine.

Perhaps the CI/CD/build problem should be addressed in a different issue/PR. That's probably not directly related to the strict mode issue.

Do you want to remove the draft state of this PR and ping Netanel?

Den-dp:
Thanks for your help @qnku @nunoarruda!

I've converted this Draft into PR, updated the description with your feedback, and also provided a CI fix.

----

Initially, I wanted to find out what is the best way to repro the issue using this repo, so that acceptance criteria for #870 would be more straightforward and obvious.

The only solution I came up with is to convert the `ng-playground` app from this repo into strict and switch it to use compiled libs from `dist` (the last one helps us by not obliging to fix all akita strict issues but only d.ts related).

But the `ng-playground` is big too and requires more time so I will do it in the next PR

PR #1055, opened 2023-02-12
refactor(general): use relative import

Den-dp:
## PR Checklist

## PR Type

```
```

The issue can be experienced only when switching your tsconfig.base paths to use `dist` instead of `src` (that is what I have tried to do in this repo while working on #1050), because in such case `jest` will not be able to import it (so this test fails)

## What is the new behavior?

Using relative paths within the same project prevents circular self-dependency

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/akita/src/__tests__/env.spec.ts
+++ b/packages/akita/src/__tests__/env.spec.ts
@@ -1,4 +1,4 @@
-import { enableAkitaProdMode } from '@datorama/akita';
+import { enableAkitaProdMode } from '../lib/env';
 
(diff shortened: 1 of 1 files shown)


NetanelBasal:
@Den-dp the build is failing.

Den-dp:
@NetanelBasal sure, was waiting for https://github.com/salesforce/akita/pull/1050/commits/5610f559b38c96623d630d525c1c354660ba68b1 to be merged.

I've rebased it and now it's green

